PR #104, opened 2022-03-15
Add: pre run feature to collecting data that will elsewhere reused

Dexus:
**What**:

Add a PRE RUN feature, that can collect data and add information to the `kwargs['<name>']` arguments for the default checks.

Currently I'm not sure how to check the new PRE RUN, maybe @yourself have an idea.

**Why**:

we need to collect some data, to reduce the time that we need run each file again and again and again ... to check against duplicated oid's and other stuff.

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation


troubadix/plugin.py
```diff
@@ -46,6 +45,19 @@ class Plugin(ABC):
...
 
 
+class PreRunPlugin(Plugin):
+    """A plugin that only runs PreRun collectors"""
+
+    @staticmethod
+    @abstractmethod
+    def run(
+        pre_run_data: dict,
+        nasl_files: List[Path],
```
bjoernricks, line 55:
```suggestion
        nasl_files: Iterable[Path],
```

Dexus: Does this match `List`? If not, it don't work here.

Dexus: @y0urself

y0urself: List are Iterables but not Iterators, thus it is okay here.

troubadix/helper/helper.py
```diff
@@ -33,18 +48,20 @@ def is_ignore_file(
     return False
 
 
+def get_path_from_root(file_name: Path, root: str = "nasl/"):
+    return str(file_name).split(root, maxsplit=1)[-1]
```
bjoernricks, line 52: IMHO this is very ugly. Converting a Path into a String to extract some Path. Doesn't feel right. Maybe https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parts can be used instead?

y0urself: So I split it into the parts and concatenate them again and before that I need to get all tuple elements that are behind the element I am looking for?

y0urself:
```python
>>> x = Path("/foo/bar/baz/nasl/bee/boop/bab.bing.nasl")
>>> x
PosixPath('/foo/bar/baz/nasl/bee/boop/bab.bing.nasl')
>>> x.parts
('/', 'foo', 'bar', 'baz', 'nasl', 'bee', 'boop', 'bab.bing.nasl')
```

There is no way to find the `root` (the `'nasl'` part in this case) from that tuple other than iterating through all parts and check each part, or am I wrong?

bjoernricks: Maybe a test for getting the exact use case would help. My problem is that a path is not a string. So I would even change the second parameter to be a path.

y0urself: e.g. like this:
```python
>>> for i, part in enumerate(x.parts):
...   if part == "nasl":
...     y = x.parts[i:]
...
>>> y
('nasl', 'bee', 'boop', 'bab.bing.nasl')
```

y0urself: I have just refactored the code, but TBH I do not think that we need to store only the part of the path in the dict.

But I am somehow a fan of not printing out the whole path from the root (`/`), since these pathes can get very long and thus unreadable ...

y0urself: > My problem is that a path is not a string.

When you want only a part of the path you need to convert it to a string. You can only have relative path or absolute path. And a `Path` is a complex structure for something that in the end is just a string. Like typing `cd /foo/bar/baz` ... A path can always be represented in a string and nearly everyone can tell you "ahh that string is a path" ... thus I am not with you in this case.
Yes a `Path` is not a string, but a path is.

bjoernricks: A path just looks like a string but it has extra meanings. You of course can convert between them.

Nevertheless maybe some other Path API fits better here for example https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.root or https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to

y0urself: > https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to

can only be used with `PurePath`...

y0urself (approved):

PR #111, opened 2022-03-23
Add: a plugin results statistic

y0urself:
**What**:

* Default: Summary will be printed after run
  * Will not run if `--no-statistic` argument will used
* Prints the statistic how many errors per plugin have been found   

Summary printing:
```
  Plugin                                   Error Count
  ----------------------------------------------------
× check_updated_date_version                       830
× check_includes                                     8
× check_todo_tbd                                     2
  ----------------------------------------------------
× sum                                              840
```

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation


troubadix/runner.py
```diff
@@ -99,6 +120,32 @@ def _report_bold_info(self, message: str):
     def _report_ok(self, message: str):
         self._term.ok(message)
 
+    def _process_plugin_results(self, results: OrderedDict):
+
```
bjoernricks, line 124: IMHO it is important to specify the internal types of the Dict. And also the method doesn't care about the OrderedDict "interface" here and just expects a dict. Btw. dicts are ordered too nowadays.
```suggestion
    def _process_plugin_results(self, results: Dict[str, List[LinterMessage]]) -> None:
```

bjoernricks (changes requested):

troubadix/runner.py
y0urself: You have introduced the `OrderedDict` in `FileResults` ;)

Dexus (approved):

Dexus (approved):

troubadix/runner.py
```diff
@@ -84,21 +105,48 @@ def _report_results(self, results: List[LinterMessage]):
...
+            plugin_results,
+        ) in results.items():
+            if plugin_results[0] or self.debug:
+                self._report_info(f"Running plugin {plugin_name}")
+
+            # add the results to the statistic
+            self.result_counts.add_result_counts(plugin_name, plugin_results[1])
+
+            with self._term.indent():
+                self._report_results(plugin_results[0])
```
bjoernricks, line 138:
```suggestion
            if plugin_result or self.debug:
                self._report_info(f"Running plugin {plugin_name}")

            # add the results to the statistic
            self.result_counts.add_result_counts(plugin_name, len(plugin_results))

            with self._term.indent():
                self._report_results(plugin_results)
```

bjoernricks (changes requested):
We should only add the result counts where it is actually used.

troubadix/runner.py
```diff
@@ -58,6 +59,19 @@ def add_plugin_results(
...
+    def __init__(self):
+        self.result_counts = {}
+
+    def add_result_counts(self, plugin: str, count: int):
+        """Add the number of results (count) to the dict for the
+        plugin name (plugin)"""
+        if count > 1:
+            if plugin not in self.result_counts:
+                self.result_counts[plugin] = 0
+            self.result_counts[plugin] += count
```
bjoernricks, line 72:
```suggestion
    def __init__(self):
        self.result_counts = defaultdict(int)

    def add_result_counts(self, plugin: str, count: int):
        """Add the number of results (count) to the dict for the
        plugin name (plugin)"""
        if count > 1:
            self.result_counts[plugin] += count
```

bjoernricks (changes requested):
Simplified ResultCounts class by using [defaultdict](https://docs.python.org/3.9/library/collections.html#collections.defaultdict) from collections module.

troubadix/runner.py
```diff
@@ -84,21 +100,50 @@ def _report_results(self, results: List[LinterMessage]):
...
+    def _process_plugin_results(
+        self, results: Dict[str, List[LinterMessage]]
+    ) -> None:
+        # print the files plugin results
+        for (
+            plugin_name,
+            plugin_results,
+        ) in results.items():
+            if plugin_results or self.debug:
+                self._report_info(f"Running plugin {plugin_name}")
```
bjoernricks, line 127:
```suggestion
                self._report_info(f"Results for plugin {plugin_name}")
```

bjoernricks (approved):

bjoernricks (approved):

PR #112, opened 2022-03-24
Add: Option to print into a log file and verbosity mechanism (-vv) instead of debug

y0urself:
**What**: Add log file support and replace --debug with verbose level 0 -2

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation


troubadix/runner.py
```diff
@@ -86,18 +94,26 @@ def _report_results(self, results: List[LinterMessage]):
 
     def _report_warning(self, message: str):
         self._term.warning(message)
+        message = message.replace("\n", "\n\t\t")
+        self._log_append(f"\t\t{message}")
```
y0urself, line 98:
```suggestion
        self._log_append(f"\t\t{message}".replace("\n", "\n\t\t"))
```

troubadix/runner.py
```diff
@@ -86,18 +94,26 @@ def _report_results(self, results: List[LinterMessage]):
...
 
     def _report_error(self, message: str):
         self._term.error(message)
+        message = message.replace("\n", "\n\t\t")
+        self._log_append(f"\t\t{message}")
```
y0urself, line 103:
```suggestion
        self._log_append(f"\t\t{message}".replace("\n", "\n\t\t"))
```

troubadix/runner.py
```diff
@@ -86,18 +94,26 @@ def _report_results(self, results: List[LinterMessage]):
...
+        self._log_append(f"\t{message}")
 
     def _report_bold_info(self, message: str):
         self._term.bold_info(message)
+        self._log_append(f"\n\n{message}")
 
     def _report_ok(self, message: str):
         self._term.ok(message)
+        message = message.replace("\n", "\n\t\t")
+        self._log_append(f"\t\t{message}")
```
y0urself, line 116:
```suggestion
        self._log_append(f"\t\t{message}".replace("\n", "\n\t\t"))
```

troubadix/runner.py
```diff
@@ -67,14 +67,22 @@ def __init__(
...
+        self._log_file = log_file
         self.special_tag_pattern = SpecialScriptTagPatterns()
         self.tag_pattern = ScriptTagPatterns()
 
+    def _log_append(self, message: str):
+        if not self._log_file:
+            return
+        with self._log_file.open(mode="a", encoding="utf-8") as f:
+            f.write(f"{message}\n")
+
```
y0urself, line 85:
```suggestion
        if self._log_file:
            with self._log_file.open(mode="a", encoding="utf-8") as f:
                f.write(f"{message}\n")

```

This would also work.

I wonder if we should print on the console, too, if a logfile is passed.

I think I would love to implement something like [this](https://docs.python.org/3/library/argparse.html#action) for verbosity. :D

```python
>>> parser = argparse.ArgumentParser()
>>> parser.add_argument('--verbose', '-v', action='count', default=0)
>>> parser.parse_args(['-vvv'])
Namespace(verbose=3)
```

pascalholthaus: Maybe move " if self._log_file " to the _report_* functions?

y0urself: mhh. this would increase code duplication ... I think i is okayish this way

pascalholthaus: verbose mode added

troubadix/runner.py
pascalholthaus: done

troubadix/runner.py
pascalholthaus: done

troubadix/runner.py
pascalholthaus: done

bjoernricks (approved):

pascalholthaus (approved):

PR #116, opened 2022-03-24
Pattern improvements

bjoernricks:
**What**:

Rework usage of common patterns

**Why**:

Make the Plugin API clean and simple as possible.

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/tests/helper/test_patterns.py
+++ b/tests/helper/test_patterns.py
@@ -1,0 +1,59 @@
+# Copyright (C) 2022 Greenbone Networks GmbH
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import unittest
+
+from troubadix.helper.patterns import (
+    ScriptTag,
+    SpecialScriptTag,
+    get_script_tag_pattern,
+    get_script_tag_patterns,
+    get_special_script_tag_pattern,
+    get_special_script_tag_patterns,
+)
+
+
+class ScriptTagPatternsTestCase(unittest.TestCase):
+    def test_same_patterns_instance(self):
+        patterns1 = get_script_tag_patterns()
+        patterns2 = get_script_tag_patterns()
+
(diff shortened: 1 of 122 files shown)


y0urself:
For `require_ports` and `require_udp_ports` I have added the following regex value pattern: https://regex101.com/r/3kx0jO/1
For `xref` it should be fine with: https://regex101.com/r/ynfZ0L/5

y0urself:
We are still very slow. I ran this locally and it took nealy 40min.
```
ℹ Time elapsed: 0:37:52.903985
  Plugin                                   Error Count
  ----------------------------------------------------
× check_updated_date_version                    158740
× check_script_version_and_last_modification_tags       28154
× check_dependencies                               185
× check_includes                                  2153
× check_copyright_text                            1076
× check_script_copyright                          1076
× check_using_display                              130
× check_todo_tbd                                   142
× check_get_kb_on_services                           6
× check_grammar                                     23
× check_badwords                                    42
× check_prod_svc_detect_in_vulnvt                   17
× check_misplaced_compare_in_if                    322
× check_copyright_year                             218
× check_forking_nasl_funcs                           8
  ----------------------------------------------------
× sum                                           192292
```

I have increased the spacing between the columns because of very long check names.

tests/helper/test_patterns.py
```diff
@@ -0,0 +1,59 @@
...
+        patterns1 = get_script_tag_patterns()
+        patterns2 = get_script_tag_patterns()
+
+        self.assertIs(patterns1, patterns2)
+
+    def test_same_pattern_instance(self):
+        pattern1 = get_script_tag_pattern(ScriptTag.AFFECTED)
+        pattern2 = get_script_tag_pattern(ScriptTag.AFFECTED)
+
+        self.assertIs(pattern1, pattern2)
```
y0urself, line 41: The init will be done for every thread, AFAIK.

tests/plugins/__init__.py
```diff
@@ -17,15 +17,6 @@
...
-)
-
 
 class PluginTestCase(unittest.TestCase):
-    def setUp(self) -> None:
-        self.tag_pattern = ScriptTagPatterns().pattern
-        self.special_tag_pattern = SpecialScriptTagPatterns().pattern
-
-        return super().setUp()
+    pass
```
y0urself, line 22: Pretty obsolete Class now ... Remove it again?

y0urself (approved):
Approving your changes. Disabled auto-merge because of my changes. You can merge, if you are okay with them.

tests/plugins/__init__.py
bjoernricks: Yes at the moment. But I have some things in mind for making the plugin tests easier to write and therefore I kept the class.

y0urself: Ok cool!

tests/helper/test_patterns.py
bjoernricks: No, threads share memory. If you mean the processes from the Pool then the answer is different. When forking a process the already loaded data gets "copied" into the new process which is "copy on write" on Linux. Therefore the init is run before the forking. But at least is is only created once per process.

PR #117, opened 2022-03-24
Raise exception if no plugin is run

bjoernricks:
**What**:

Fail if no plugin is run.

**Why**:

This is very likely an error on the users side for example a typo in the
include parameter.

**How**:

**Checklist**:

- [x] Tests
- [ ] Documentation

### DIFF
--- a/tests/test_runner.py
+++ b/tests/test_runner.py
@@ -25,7 +25,9 @@ from pontos.terminal.terminal import Terminal
 
 from troubadix.plugin import LinterError, LinterResult
 from troubadix.plugins import _NASL_ONLY_PLUGINS
-from troubadix.runner import Runner
+from troubadix.runner import Runner, TroubadixException
+
+_here = Path(__file__).parent
 
 
 class TestRunner(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


tests/test_runner.py
```diff
@@ -247,7 +249,19 @@ def test_runner_run_ok_without_debug(self):
 
         output = f.getvalue()
         self.assertIn(f"Checking {nasl_file}", output)
-        self.assertNotIn("Running plugin check_missing_desc_exit", output)
+        self.assertNotIn("Results for plugin check_missing_desc_exit", output)
```
y0urself, line 252: Why? The output is always 
```
ℹ Checking common2/2008/gb_w3c_amaya_mult_bof_vuln_dec08_lin.nasl (413/415)
ℹ     Running plugin check_openvas_lint
×         openvas-nasl-lint not found within your PATH. Please install 'openvas' and make sure it is available within your PATH.
ℹ     Running plugin check_updated_date_version
×         Changed VT has a not updated script_version();
```
Or not?

```suggestion
        self.assertNotIn("Running plugin check_missing_desc_exit", output)
```

y0urself (changes requested):

tests/test_runner.py
bjoernricks: Because actually the plugin has run already and therefore the statement is technically wrong

y0urself: Yes that is correct. But it is not enough to change this in a single assertion within a test. :P 
Should we change it all over, or leave it?

bjoernricks: Oh sorry this change should not have been included here. I'll fix that.

y0urself (approved):

PR #120, opened 2022-03-25
Change console output to list results of a plugin

bjoernricks:
**What**:

Change console output to list results of a plugin

**Why**:

Actually the plugin has been run already. We just collected and
displayed the results.

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/runner.py
+++ b/troubadix/runner.py
@@ -141,8 +141,10 @@ class Runner:
             plugin_name,
             plugin_results,
         ) in results.items():
-            if plugin_results or self.debug:
+            if plugin_results:
                 self._report_info(f"Results for plugin {plugin_name}")
+            elif self.debug:
+                self._report_ok(f"No results for plugin {plugin_name}")
 
             # add the results to the statistic
             self.result_counts.add_result_counts(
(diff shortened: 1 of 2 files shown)


troubadix/runner.py
```diff
@@ -126,7 +126,7 @@ def _process_plugin_results(
             plugin_results,
         ) in results.items():
             if plugin_results or self.debug:
-                self._report_info(f"Running plugin {plugin_name}")
+                self._report_info(f"Results for plugin {plugin_name}")
```
y0urself, line 129: We could also make this even better. How about writing `f"No results for plugin {plugin_name}"`

```python
if plugin_results:
  self._report_info(f"Results for plugin {plugin_name}")
elif self.debug:
  f"No results for plugin {plugin_name}"
``` 
  
What do you think?

y0urself (changes requested):

troubadix/runner.py
bjoernricks: Would be nice 👍🏻

y0urself (dismissed):

y0urself (approved):

PR #125, opened 2022-03-28
Remove: plugin openvas_lint.py

gb-friotte:
**What**:

  The plugin openvas_lint.py and its respective test are removed. 

**Why**:

We will run openvas-nasl-lint directly within an GitHub Action workflow and do not use this within troubadix

**How**:

  - [x] checked if helper.get_root is still needed
  - [x] checked if helper.subprocess_cmd is still needed

**Checklist**:

- [ ] Tests
- [ ] Documentation


y0urself (approved):
Thx!

PR #136, opened 2022-04-04
Let the description check find everything

pascalholthaus:
**What**: Let the description check find everything.

**Why**:
To be on the safe side, since script_description
does not exist anymore.

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/description.py
+++ b/troubadix/plugins/description.py
@@ -38,7 +38,7 @@ class CheckDescription(LineContentPlugin):
             file_content: The content of the file that is going to be
                           checked
         """
-        pattern = re.compile(r"script_description\(.+\);", re.IGNORECASE)
+        pattern = re.compile(r"script_description", re.IGNORECASE)
 
         for line in lines:
             match = pattern.search(line)
--- a/troubadix/plugins/description.py
+++ b/troubadix/plugins/description.py
@@ -17,18 +17,18 @@
 
 import re
 from pathlib import Path
-from typing import Iterable, Iterator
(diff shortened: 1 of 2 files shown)


troubadix/plugins/description.py
```diff
@@ -38,7 +38,7 @@ def run(
             file_content: The content of the file that is going to be
                           checked
         """
-        pattern = re.compile(r"script_description\(.+\);", re.IGNORECASE)
+        pattern = re.compile(r"script_description", re.IGNORECASE)
```
y0urself, line 41: why?
What about `pattern = re.compile(r"script_description\s*\(\s*.+\s*\)\s*;", re.IGNORECASE)`
Or what about just check `if "script_description" in content` (make it a `FileContentPlugin`)?

pascalholthaus: Thought it would be nice to have a line.
if "script_description" in content is better.

troubadix/plugins/description.py
```diff
@@ -38,12 +38,10 @@ def run(
...
-            match = pattern.search(line)
-            if match:
-                yield LinterError(
-                    "VT/Include is using deprecated 'script_description':"
-                    f" {line}"
-                )
+        script_description = re.compile(
+            r"script_description", re.IGNORECASE
+        ).search(file_content)
+        if script_description:
```
y0urself, line 44:
```suggestion
        if "script_description" in file_content:
```

y0urself (approved):

PR #140, opened 2022-04-07
Init error count

pascalholthaus:
**What**: Basic error and warning count

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/runner.py
+++ b/troubadix/runner.py
@@ -80,12 +80,20 @@ class FileResults:
 class ResultCounts:
     def __init__(self):
         self.result_counts = defaultdict(int)
+        self.error_count = 0
+        self.warning_count = 0
 
     def add_result_counts(self, plugin: str, count: int):
         """Add the number of results (count) to the dict for the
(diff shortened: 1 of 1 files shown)


troubadix/runner.py
```diff
@@ -194,8 +195,15 @@ def _report_statistic(self) -> None:
             overall += count
             self._term.error(f"{plugin:50} {count:11}")
         self._term.print("-" * 62)
+        self._term.error(f"{'err':50} {self._error_count:11}")
```
y0urself, line 198:
```suggestion
        self._term.error(f"{'err':50} {self._error_count:11}")
        self._term.warning(f"{'war':50} {self._warning_count:11}")
```

troubadix/runner.py
```diff
@@ -194,8 +196,18 @@ def _report_statistic(self) -> None:
             overall += count
```
y0urself, line 196:
```suggestion
```

PR #147, opened 2022-04-08
Add: Plugin for handling of missing solutions added

gb-friotte:
**What**:

Added plugin to evaluate missing solutions within solution tag in collaboration with @pascalholthaus 

**Why**:

FE-173

**How**:

**Checklist**:

- [x] Tests
- [ ] Documentation

### DIFF
--- a/tests/plugins/test_files/nasl/21.04/fail_solution_template.nasl
+++ b/tests/plugins/test_files/nasl/21.04/fail_solution_template.nasl
@@ -1,0 +1,52 @@
+###############################################################################
+# OpenVAS Vulnerability Test
+#
+# HP Comware Devices Detect (SNMP)
+#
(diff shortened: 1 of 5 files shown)


Dexus:
@y0urself this is the example why I was adding the PreRunner Plugin in a different style - which is now the same stuff as a normal plugin and we run over each files multiple times again...

y0urself:
> @y0urself this is the example why I was adding the PreRunner Plugin in a different style - which is now the same stuff as a normal plugin and we run over each files multiple times again...

The way I implemented the `DuplicateOID` Plugin is still one time over all files. But for each file it runs through the whole `dict`, which makes it kind of O(n^2).

bjoernricks (approved):

PR #150, opened 2022-04-11
Fix false positives in check_reporting_consistency

pascalholthaus:
**What**: Fix false positives in check_reporting_consistency

**Why**:
DEVOPS-179

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation


Dexus:
why we can't use the functions that we have?

pascalholthaus:
We have functions for security_message and log_message?

bjoernricks (approved):

PR #163, opened 2022-04-13
Improvements

bjoernricks:
**What**:

* Don't hardcode encoding
* Unify result reporting
* Remove obsolete code
* Improve Plugins handling
* Some cleanup
* Remove unused arguments from troubadix

**Why**:

**How**:

**Checklist**:

- [x] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/duplicate_oid.py
+++ b/troubadix/plugins/duplicate_oid.py
@@ -20,6 +20,7 @@ from pathlib import Path
 from typing import Iterator, List
 
 from troubadix.helper import (
+    CURRENT_ENCODING,
     SpecialScriptTag,
     get_special_script_tag_pattern,
     get_path_from_root,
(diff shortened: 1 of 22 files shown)


troubadix/plugins/duplicate_oid.py
```diff
@@ -54,43 +55,46 @@ def run(
...
+            # search for deprecated script_id
+            match = get_special_script_tag_pattern(SpecialScriptTag.OID).search(
+                content
+            )
+
+            if match:
+                oid = match.group("oid")
+            else:
                 match = get_special_script_tag_pattern(
-                    SpecialScriptTag.OID
```
Dexus, line 63: why change it? but not use later the correct mach?

```
oid = OPENVAS_OID_PREFIX + match.group("id")
```

Dexus: I think it is already a typo that there was used `oid` in the old code.

bjoernricks: I didn't change anything regarding the content of this code. Just the if clause to remove the else branch.

Dexus (approved):

PR #167, opened 2022-04-14
Plugin infrastructure

bjoernricks:
**What**:

* Introduce state to the plugins
* Pass a context to the plugins

**Why**:

Having a state will allow for having different phases in plugins. In our case we can split updating things into a different phase later on.

**How**:

**Checklist**:

- [x] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/double_end_points.py
+++ b/troubadix/plugins/double_end_points.py
@@ -16,20 +16,18 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import re
-from pathlib import Path
+
 from typing import Iterator
 
 from troubadix.helper.patterns import get_common_tag_patterns
-from troubadix.plugin import FileContentPlugin, LinterError, LinterResult
+from troubadix.plugin import LinterError, LinterResult, FilePlugin
 
 
-class CheckDoubleEndPoints(FileContentPlugin):
+class CheckDoubleEndPoints(FilePlugin):
     name = "check_double_end_points"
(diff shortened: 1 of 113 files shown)


troubadix/plugins/double_end_points.py
```diff
@@ -16,20 +16,18 @@
...
+
 from typing import Iterator
 
 from troubadix.helper.patterns import get_common_tag_patterns
-from troubadix.plugin import FileContentPlugin, LinterError, LinterResult
+from troubadix.plugin import LinterError, LinterResult, FilePlugin
 
 
-class CheckDoubleEndPoints(FileContentPlugin):
+class CheckDoubleEndPoints(FilePlugin):
```
y0urself, line 26: Why are you now mixing up using `FilePlugin` and `LineContentPlugin`s?
Why not use `FileContentPlugin` for these?

bjoernricks: `FilePlugin`s are plugins acting on files and getting a `FilePluginContext`. `LineContentPlugin` and `FileContentPlugin` are more or less legacy plugins that are now derived from `FilePlugin` and could also be replaced by a `FilePlugin`. The `check_content` and `check_lines` are convenient methods to allow to keep the old behavior. I've replaces `FileContentPlugin`s with a `FilePlugin` where the `nasl_file` argument wasn't used at all.

Dexus (approved):
for me looks all reasonable and ok

PR #190, opened 2022-04-25
Add --fix argument

bjoernricks:
**What**:

Introduce an explicit fix phase for plugins and a `--fix` CLI argument to run this phase.

https://jira.greenbone.net/browse/DEVOPS-208

**Why**:

Separate checks and fixing content. This is how other formatters and linters are working.

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/copyright_year.py
+++ b/troubadix/plugins/copyright_year.py
@@ -47,7 +47,6 @@ class CheckCopyrightYear(LineContentPlugin):
         if nasl_file.suffix == ".inc":
             return
 
-        report = ""
         creation_date = ""
         creation_year = ""
         copyrights = []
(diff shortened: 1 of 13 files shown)


troubadix/plugins/copyright_year.py
```diff
@@ -70,20 +70,20 @@ def check_lines(
...
-                report += f"\n{key.strip()}\n"
+        for nr, line, copyright_year in copyrights:
+            if copyright_year != creation_year:
+                report += f"Line {nr}: {line.strip()}\n"
 
         if len(report) > 0:
             yield LinterError(
                 "VT contains a Copyright year not matching "
-                f"the year in {copyright_year} at the following lines:\n"
+                f"the year in {creation_year} at the following lines:\n"
```
y0urself, line 87: IMHO we should print an Error for each found line individually here, like everywhere else ...
```python
        for nr, line, copyright_year in copyrights:
              if copyright_year != creation_year:
                  yield LinterError(
                     f"VT contains a Copyright year not matching"
                     f"the year in {creation_year} (line {nr})"
                  )
```

bjoernricks: Yes you are right. I'll change that.

bjoernricks: Done with https://github.com/greenbone/troubadix/pull/190/commits/18de06c062f609d33bc69798ac01ce3e54abfcbd

y0urself (approved):

PR #193, opened 2022-04-26
Plugin http_links_in_tags overhauled check_to_continue method

mbrinkhoff:
**What**:

Reworked check_to_continue method to be easier to maintain by creating a simple list of text phrases to exclude.

**Why**:

It is easier to maintain and add new entries

**How**:

Passes the existing unittests

**Checklist**:

- [x] Tests
- [ ] Documentation


y0urself (commented):
Would be awesome to write a test for each exclusion/a single test that iterates through all the exclusions.

mbrinkhoff:
Added testcases

bjoernricks (approved):

PR #221, opened 2022-05-18
Change: Simplify valid_oid.py plugin. Drop invalid comment.

cfi-gb:
**What**:

- This was implemented quite sub-optimal in the old Feed-QA in the past but got improved within it in the meantime. Applying the same changes / improvements to Troubadix as well
- Also dropped one invalid comment about some badword checks
- One test for an "unknown" vendor has been changed from 14 (which is now Rocky Linux) to 99 so that we don't need to change that one after every addition of a new Linux distro

**Why**:

Easier maintenance of additions to this code.

**How**:

```
$ troubadix --include-tests check_valid_oid -f
```

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/valid_oid.py
+++ b/troubadix/plugins/valid_oid.py
@@ -308,7 +308,7 @@ class CheckValidOID(FileContentPlugin):
             elif vendor_number == "14":
                 if family != f"Rocky Linux {family_template}":
                     yield LinterError(
-                        f"script_oid() {is_using_reserved} Rocky Linux VTs "
+                        f"script_oid() {is_using_reserved} Rocky Linux "
                         f"'{str(oid)}'",
                         file=nasl_file,
                         plugin=self.name,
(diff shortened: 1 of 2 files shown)


troubadix/plugins/valid_oid.py
```diff
@@ -122,8 +146,10 @@ def check_content(
                     )
                 return
 
-            # Fixed OID-scheme for SUSE SLES OS OIDs
-            elif "1.3.6.1.4.1.25623.1.1.4." in oid:
+            # TBD: 3 is n/a in Confluence
+            # elif vendor_number == "3":
```
y0urself, line 150: Maybe already throw a warning, that an OID with a probably not available vendor has been found?

y0urself (changes requested):

troubadix/plugins/valid_oid.py
cfi-gb: Mhhh, indeed we could just drop that for now until it is clarified what vendor should be using this OID range.

```suggestion
```

n-thumann (approved):

y0urself (approved):

PR #222, opened 2022-05-18
Fix: Exclusion in valid_oid.py

cfi-gb:
**What**:

Running the plugin shows the following previously:

```
$ troubadix --include-tests check_valid_oid --file nasl/common/gsf/2018/vmware/gb_vmware_fusion_vmxnet3_stack_memory_usage_vuln_macosx.nasl -v
ℹ Start linting 1 files ... 
/path/to/nasl/common/gsf/2018/vmware/gb_vmware_fusion_vmxnet3_stack_memory_usage_vuln_macosx.nasl
ℹ Checking gsf/2018/vmware/gb_vmware_fusion_vmxnet3_stack_memory_usage_vuln_macosx.nasl (1/1)
ℹ     Results for plugin check_valid_oid
×         script_oid() is using an invalid OID '1.3.6.1.4.1.25623.1.0.8141330' (unassigned OID range)
ℹ Time elapsed: 0:00:00.015437
× check_valid_oid                                         1        0
ℹ sum                                                     0        1
```

This happened because of a `,` which has been missed to be added in #160

**Why**:

Exclude an error for an OID which already exists since longer time in the feed and which can't be changed.

**How**:

Run the previously shown command before / after this MR

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/valid_oid.py
+++ b/troubadix/plugins/valid_oid.py
@@ -403,6 +403,7 @@ class CheckValidOID(FileContentPlugin):
         exceptions = [
             "ossim_server_detect.nasl",
             "gsf/2018/vmware/gb_vmware_fusion_vmxnet3_"
+            "stack_memory_usage_vuln_macosx.nasl",
             "enterprise/2018/vmware/gb_vmware_fusion_vmxnet3_"
             "stack_memory_usage_vuln_macosx.nasl",
             "2008/asterisk_sdp_header_overflow.nasl",


Dexus (approved):

troubadix/plugins/valid_oid.py
```diff
@@ -403,6 +403,7 @@ def check_content(
         exceptions = [
             "ossim_server_detect.nasl",
             "gsf/2018/vmware/gb_vmware_fusion_vmxnet3_"
+            "stack_memory_usage_vuln_macosx.nasl",
             "enterprise/2018/vmware/gb_vmware_fusion_vmxnet3_"
```
Dexus, line 407: please add on all line ends a `,`

Dexus (changes requested):

troubadix/plugins/valid_oid.py
cfi-gb: Please correct me if i'm wrong but that would mean that `gsf/2018/vmware/gb_vmware_fusion_vmxnet3_` and `stack_memory_usage_vuln_macosx.nasl` would be seen as separate strings / files while it should be `gsf/2018/vmware/gb_vmware_fusion_vmxnet3_stack_memory_usage_vuln_macosx.nasl` (a single file / string).

Dexus: You're right, I just thought they were all different files and not contiguous strings.

Dexus (approved):

PR #244, opened 2022-06-01
Add allow list for duplicated script_dependencies

pascalholthaus:
**What**:Add allow list for duplicated script_dependencies

**Why**:
Note that we shouldn't generally allow script_dependencies() multiple times (because this might be an error due to copy'n'paste or similar) but only for the existing exclusions AND the three VTs mentioned above (because they are some kind of a special case which is allowed for now because we know that the tag isn't used wrongly).

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/tests/plugins/test_duplicated_script_tags.py
+++ b/tests/plugins/test_duplicated_script_tags.py
@@ -101,8 +101,8 @@ class CheckDuplicatedScriptTagsTestCase(PluginTestCase):
     def test_excluded_dependencies(self):
         path = Path("gsf/PCIDSS/v2.0/PCI-DSS-2.0.nasl")
         content = (
-            'script_dependencies(name:"Test", type:"checkbox");\n'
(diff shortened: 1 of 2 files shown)


tests/plugins/test_duplicated_script_tags.py
```diff
@@ -97,3 +97,33 @@ def test_excluded_tag(self):
         results = list(plugin.run())
 
         self.assertEqual(len(results), 0)
+
+    def test_excluded_dependencies(self):
+        path = Path("gsf/PCIDSS/v2.0/PCI-DSS-2.0.nasl")
+        content = (
+            'script_dependencies(name:"Test", type:"checkbox");\n'
```
cfi-gb, line 104: The syntax of `script_dependencies()` is basically `script_dependencies("vt1.nasl", "vt2.nasl");`

bjoernricks (approved):

PR #250, opened 2022-06-03
More excluded inc files

pascalholthaus:
**What**: More excluded inc files

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation


bjoernricks (approved):

y0urself (commented):
Can you add a test with an .inc file and check if it is skipped? We should probably do this for all checks ...

y0urself (approved):

PR #271, opened 2022-06-23
Fix: False Positives in the dependencies check

y0urself:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/dependencies.py
+++ b/troubadix/plugins/dependencies.py
@@ -31,6 +31,8 @@ from troubadix.plugin import (
     LinterWarning,
 )
 
+FEED_VERSIONS = ["common", "21.04", "22.04"]
+
 
 class CheckDependencies(FilePlugin):
     name = "check_dependencies"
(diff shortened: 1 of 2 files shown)


troubadix/plugins/dependencies.py
```diff
@@ -72,7 +74,9 @@ def run(
                     if "+d+.nasl" in dep:
                         continue
 
-                    if not (root / dep).exists():
+                    if not any(
+                        (root / vers / dep).exists() for vers in FEED_VERSIONS
+                    ):
```
pascalholthaus, line 79: Maybe I'm wrong, but when the feed is generated it's all in one directory called nasl. Maybe check (root / dep) too?

y0urself: i thought about this, but this script is not running there, I think?

pascalholthaus: At least on every PR it is running on a generated feed release.

troubadix/plugins/dependencies.py
```diff
@@ -31,6 +31,8 @@
     LinterWarning,
 )
 
+FEED_VERSIONS = ["common", "21.04", "22.04"]
```
y0urself, line 34:
```suggestion
FEED_VERSIONS = ["common", "21.04", "22.04", ""]
```

troubadix/plugins/dependencies.py
y0urself: I have added `""` ;)

pascalholthaus (approved):

PR #272, opened 2022-06-23
Fix: Duplicated checks in copyright_text and script_copyright

cfi-gb:
**What**:

While migrating/porting the "old" `check_copyright_text.py` to Troubadix it seems that it was tried to split the two checks in that step into two separate plugins but it was missed to remove the check from one plugin which had been covered in the other one causing duplicated results to be reported.

This PR should fix that by removing the check from the plugins which had been covered by the other plugin.

Notes:
- I have also added a few additional cases for one plugin test.
- I haven't put any additional work into the text output of `script_copyright.py` (Because e.g. it also reports if the `script_copyright()` is completely missing but this is currently not reflected in the output) because this tag will be deprecated and removed completely in the future anyway
- The unit test in `test_script_copyright.py` will most likely fail and it would be great to get some support by the @greenbone/devops team to get it updated because i'm not familiar with unit tests.

Related to: VTD-1534

**Why**:

Avoid duplicated reporting of the same error.

**How**:

Via the related / included unit tests.

**Checklist**:

- [x] Tests
- [ ] Documentation

### DIFF
--- a/tests/plugins/test_script_copyright.py
+++ b/tests/plugins/test_script_copyright.py
@@ -62,3 +62,41 @@ class CheckScriptCopyrightTestCase(PluginTestCase):
             "copyright statement.",
             results[0].message,
         )
+
+    def test_copyright_error2(self):
+        path = Path("some/file.nasl")
+        content = (
+            'script_copyright("This script is Copyright (C) 2020 Foo Bar")'
+        )
+        fake_context = self.create_file_plugin_context(
+            nasl_file=path, file_content=content
+        )
+        plugin = CheckScriptCopyright(fake_context)
+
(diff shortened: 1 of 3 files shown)


y0urself:
Would be awesome if you could use a feature branch instead of a fork as it is a lot easier to checkout your branch.

cfi-gb:
> Would be awesome if you could use a feature branch instead of a fork as it is a lot easier to checkout your branch.

Sure, will take care of that in a few minutes. Back then when i started to submit PRs i didn't had permissions to directly push to this repo but it seems the permissions are not assigned.

cfi-gb:
Replaced by #274

PR #274, opened 2022-06-24
Fix: Duplicated checks in copyright_text and script_copyright

cfi-gb:
**What**:

Replaced #272 as a feature branch.

While migrating/porting the "old" `check_copyright_text.py` to Troubadix it seems that it was tried to split the two checks in that step into two separate plugins but it was missed to remove the check from one plugin which had been covered in the other one causing duplicated results to be reported.

This PR should fix that by removing the check from the plugins which had been covered by the other plugin.

Notes:
- I have also added a few additional cases for one plugin test.
- I haven't put any additional work into the text output of `script_copyright.py` (Because e.g. it also reports if the `script_copyright()` is completely missing but this is currently not reflected in the output) because this tag will be deprecated and removed completely in the future anyway
- The unit test in `test_script_copyright.py` will most likely fail and it would be great to get some support by the @greenbone/devops team to get it updated because i'm not familiar with unit tests.

Related to: VTD-1534

**Why**:

Avoid duplicated reporting of the same error.

**How**:

Via the related / included unit tests.

**Checklist**:

- [x] Tests
- [ ] Documentation

### DIFF
--- a/tests/plugins/test_copyright_text.py
+++ b/tests/plugins/test_copyright_text.py
@@ -82,19 +82,12 @@ class CheckCopyrightTextTestCase(PluginTestCase):
         plugin = CheckCopyrightText(fake_context)
 
         results = list(plugin.run())
-        self.assertEqual(len(results), 2)
+        self.assertEqual(len(results), 1)
 
         self.assertIsInstance(results[0], LinterError)
         self.assertEqual(
-            "The VT is using an incorrect syntax for its copyright "
-            "statement. Please start (EXACTLY) with:\n"
-            "'script_copyright(\"Copyright (C) followed by the year "
-            "(matching the one in creation_date) and the author/company.",
-            results[0].message,
-        )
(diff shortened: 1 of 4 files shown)


y0urself (approved):

cfi-gb:
@y0urself Thanks a lot for taking care of the unit tests / your support :+1:

cfi-gb:
For the references, previous to this MR the output of both plugins from a Troubadix run against the whole feed looked like this:

```
2022-06-22T07:39:35.6975643Z × check_script_copyright                              1543
2022-06-22T07:39:35.6974816Z × check_copyright_text                                 1543
```

Now we have the more correct numbers due to the de-duplication of the checks:

```
2022-06-24T13:10:05.5941281Z × check_script_copyright                                 21
2022-06-24T13:10:05.5941727Z × check_copyright_text                                 1522
```

PR #280, opened 2022-06-27
Add common, 21.04, 22.04 folder to search path in vt_placement.py

pascalholthaus:
**What**: Add common, 21.04, 22.04 folder to search path in vt_placement.py and move the 
FEED_VERSIONS list to the helper module.

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/vt_placement.py
+++ b/troubadix/plugins/vt_placement.py
@@ -21,11 +21,9 @@ from pathlib import Path
 from typing import Iterator
 
 from troubadix.helper import ScriptTag, SpecialScriptTag, get_script_tag_pattern
-from troubadix.helper.helper import ENTERPRISE_FOLDERS
+from troubadix.helper.helper import ENTERPRISE_FOLDERS, FEED_VERSIONS
 from troubadix.helper.patterns import _get_special_script_tag_pattern
 from troubadix.plugin import FileContentPlugin, LinterError, LinterResult
-
-FEED_VERSIONS = ["common", "21.04", "22.04", ""]
 
 
 class CheckVTPlacement(FileContentPlugin):
--- a/troubadix/plugins/vt_placement.py
+++ b/troubadix/plugins/vt_placement.py
@@ -25,6 +25,8 @@ from troubadix.helper.helper import ENTERPRISE_FOLDERS
 from troubadix.helper.patterns import _get_special_script_tag_pattern
 from troubadix.plugin import FileContentPlugin, LinterError, LinterResult
 
+FEED_VERSIONS = ["common", "21.04", "22.04", ""]
+
 
 class CheckVTPlacement(FileContentPlugin):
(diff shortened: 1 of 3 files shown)


troubadix/plugins/vt_placement.py
```diff
@@ -25,6 +25,8 @@
 from troubadix.helper.patterns import _get_special_script_tag_pattern
 from troubadix.plugin import FileContentPlugin, LinterError, LinterResult
 
+FEED_VERSIONS = ["common", "21.04", "22.04", ""]
```
cfi-gb, line 28: As this is also in use in dependencies.py since #271 wouldn't it make sense to somehow define this more globally so that it needs to be only updated at one place once there is a new GVM version?

y0urself: @pascalholthaus he is right. Maybe add this into https://github.com/greenbone/troubadix/blob/main/troubadix/helper/helper.py, we already defined other statics there. :)

pascalholthaus: Done

y0urself (approved):

cfi-gb:
Note that the `Conventional commit message` was missed here which means that this change probably won't be listed in changelog of the next release (until the related tool is picking the changelog entry from the PR title).

y0urself:
> Note that the `Conventional commit message` was missed here which means that this change probably won't be listed in changelog of the next release (until the related tool is picking the changelog entry from the PR title).

No worries, I have adjusted the Message for the sqash merge message, thus it is a correct CC message indeed. See https://github.com/greenbone/troubadix/commit/76322a301bc4fad1f4f47a99adcc77f2077ba901

cfi-gb:
> > Note that the `Conventional commit message` was missed here which means that this change probably won't be listed in changelog of the next release (until the related tool is picking the changelog entry from the PR title).
> 
> No worries, I have adjusted the Message for the sqash merge message, thus it is a correct CC message indeed. See [76322a3](https://github.com/greenbone/troubadix/commit/76322a301bc4fad1f4f47a99adcc77f2077ba901)

Ah, indeed. I have completely missed that the squash merge message was adjusted, thanks a lot for taking care :+1:

y0urself:
> > > Note that the `Conventional commit message` was missed here which means that this change probably won't be listed in changelog of the next release (until the related tool is picking the changelog entry from the PR title).
> > 
> > 
> > No worries, I have adjusted the Message for the sqash merge message, thus it is a correct CC message indeed. See [76322a3](https://github.com/greenbone/troubadix/commit/76322a301bc4fad1f4f47a99adcc77f2077ba901)
> 
> Ah, indeed. I have completely missed that the squash merge message was adjusted, thanks a lot for taking care 👍

Sadly github does not show a "preview" of that message in the info ... :/
Bildschirmfoto 2022-06-30 um 07 54 10

PR #297, opened 2022-07-04
Remove: Handling of unsupported script_id() tag.

cfi-gb:
**What**:

Remove any special handling of `script_id()` from Troubadix.

The `script_id()` itself is gone since quite some years with greenbone/openvas-scanner#54 and running a VT using it yields the following:

```
lib  nasl-Message: 08:24:56.287: [10679](oid.nasl:1) Undefined function 'script_id'

Error while processing oid.nasl.
1 scripts with one or more errors found
```

Furthermore `script_id()` is also already included in the `deprecated_functions.py` plugin and would be also reported by that if used.

**Why**:

Remove not required code.

**How**:

N/A

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/helper/patterns.py
+++ b/troubadix/helper/patterns.py
@@ -198,7 +198,6 @@ class SpecialScriptTag(Enum):
     MANDATORY_KEYS = "mandatory_keys"
     NAME = "name"
     OID = "oid"  # script_oid("1.3.6.1.4.1.25623.1.0.XXXXXX");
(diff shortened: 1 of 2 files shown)


mbrinkhoff (approved):
Tested it and it works, script_id is catched by `deprecated_functions` plugin.

The `deprecated_functions` check could be less strict, but with openvas-nasl-lint in place also and no `script_id` left in the current VTS repo, this LGTM.

PR #298, opened 2022-07-04
Fix: Special script tag name can not be a regex in script_calls_empty_values

pascalholthaus:
**What**:
Special script tag name can not be a regex in script_calls_empty_values. This leads to regex mismatches. Further only check script_tag and script_xref in script_calls_empty_values.

**Why**:
- DEVOPS-291
- DEVOPS-286

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/script_calls_empty_values.py
+++ b/troubadix/plugins/script_calls_empty_values.py
@@ -59,13 +59,13 @@ class CheckScriptCallsEmptyValues(FileContentPlugin):
             )
 
         for call in SpecialScriptTag:
+            if call == SpecialScriptTag.ADD_PREFERENCE:
+                continue
             # Special script tag name can not be a regex
             matches = _get_special_script_tag_pattern(
                 name=call.value, value=""
             ).finditer(file_content)
             for match in matches:
-                # if "script_add_preference" in match.group(0):
-                #    continue
                 yield LinterError(
                     f"{match.group(0)} does not contain a value",
                     file=nasl_file,
--- a/troubadix/plugins/script_calls_empty_values.py
+++ b/troubadix/plugins/script_calls_empty_values.py
@@ -61,7 +61,6 @@ class CheckScriptCallsEmptyValues(FileContentPlugin):
         for call in SpecialScriptTag:
             if call == SpecialScriptTag.ADD_PREFERENCE:
                 continue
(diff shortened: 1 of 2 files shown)


troubadix/plugins/script_calls_empty_values.py
```diff
@@ -57,12 +58,16 @@ def check_content(
...
-        for match in matches:
-            yield LinterError(
-                f"{match.group(0)} does not contain a value",
-                file=nasl_file,
-                plugin=self.name,
-            )
+        for call in SpecialScriptTag:
+            if call == SpecialScriptTag.ADD_PREFERENCE:
+                continue
+            # Special script tag name can not be a regex
```
cfi-gb, line 64:
```suggestion
```

IMHO this could be also completely dropped.

troubadix/plugins/script_calls_empty_values.py
```diff
@@ -57,12 +58,16 @@ def check_content(
...
-        matches = _get_special_script_tag_pattern(
-            name=r"(?!add_preferences).*", value=""
-        ).finditer(file_content)
-        for match in matches:
-            yield LinterError(
-                f"{match.group(0)} does not contain a value",
-                file=nasl_file,
-                plugin=self.name,
-            )
+        for call in SpecialScriptTag:
```
cfi-gb, line 61: AFAICT the `SpecialScriptTag` Enum contains various script_tags like e.g. the following:

```
script_name("foo");
```

which don't need to be checked by this plugin at all because they don't have a `value:` parameter.

Unfortunately i don't have a full list of functions which needs to be checked but we might want to ask the scanner team for such one.

troubadix/plugins/script_calls_empty_values.py
pascalholthaus: Removed

troubadix/plugins/script_calls_empty_values.py
cfi-gb: I will move this into an improvement ticket because the original issue seems to be already solved.

cfi-gb (commented):
LGTM but needs an additional review by some one more familiar with Python.

bjoernricks (approved):

PR #299, opened 2022-07-04
Fix: --from-file and --files, auto remove duplicate files

pascalholthaus:
**What**: Auto remove duplicate files from --from-file and --files file list.

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/troubadix.py
+++ b/troubadix/troubadix.py
@@ -94,7 +94,7 @@ def generate_patterns(
     return include_patterns, exclude_patterns
 
 
-def from_file(include_file: Path, term: Terminal) -> Iterable[Path]:
+def from_file(include_file: Path, term: Terminal) -> List[Path]:
     """Parse the given file containing a list of files into"""
     try:
         # Return file list and remove duplicate files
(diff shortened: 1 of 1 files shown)


y0urself (commented):
Hacky hack! :)

y0urself (approved):

troubadix/troubadix.py
```diff
@@ -97,10 +97,17 @@ def generate_patterns(
...
+        return list(
+            set(
+                [
+                    Path(f)
+                    for f in include_file.read_text(
+                        encoding="utf-8"
+                    ).splitlines()
+                ]
+            )
+        )
```
bjoernricks, line 110: Just return the set here. A set is also an Iterable

bjoernricks (changes requested):

troubadix/troubadix.py
pascalholthaus: I get some errors with it.
  File "/home/work/work/troubadix/troubadix/troubadix.py", line 162, in main
    first_file = files[0].resolve()
TypeError: 'set' object is not subscriptable

bjoernricks: Ok in that case the typing of the return value of the `from_file` function should be changed to `List` too

pascalholthaus: Ah sorry missed it.

bjoernricks (dismissed):

PR #300, opened 2022-07-04
Fix: --from-file and --files, auto remove duplicate files

pascalholthaus:
**What**:
Fix: --from-file and --files, auto remove duplicate files

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation


cfi-gb:
This had replaced #299

PR #306, opened 2022-07-06
Add: URL allow list for urls not recognized by the validators api

pascalholthaus:
**What**:
Add URL allow list for urls not recognized by the validators api.
If anyone has a better solution let me know.

**Why**:
Some urls are not recognized by the validators api.

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation


bjoernricks (approved):
Let's use this hack for now.

PR #308, opened 2022-07-06
Fix: Don't run last_modification.py against .inc files.

cfi-gb:
**What**:

With #304 the plugin got reworked to always run by default. But as `.inc` files don't have such a `last_modification` date this change causes false positives / errors against such `.inc` files.

**Why**:

Fix false positives

**How**:

Run e.g. the following below. Before this PR Troubadix reported a false positive which should be gone afterwards.

```
$ troubadix --files nasl/common/ssl_funcs.inc -v
ℹ Start linting 1 files ... 
ℹ Checking common/ssl_funcs.inc (1/1)
ℹ     Results for plugin check_last_modification
×         VT does not contain a modification day script tag.
ℹ     Results for plugin check_todo_tbd
⚠         VT contains #TODO/TBD/@todo keywords.
⚠         VT contains #TODO/TBD/@todo keywords.
⚠         VT contains #TODO/TBD/@todo keywords.
⚠         VT contains #TODO/TBD/@todo keywords.
ℹ Time elapsed: 0:00:05.399348
× check_last_modification                                 1        0
⚠ check_todo_tbd                                          0        4
ℹ sum                                                     1        4
```

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/last_modification.py
+++ b/troubadix/plugins/last_modification.py
@@ -29,6 +29,9 @@ class CheckLastModification(FilePlugin):
     name = "check_last_modification"
 
     def run(self) -> Iterator[LinterResult]:
+        if self.context.nasl_file.suffix == ".inc":
+            return
+
         self.new_file_content = None
 
         # update modification date


Orphx (dismissed):

troubadix/plugins/last_modification.py
```diff
@@ -96,6 +99,9 @@ def run(self) -> Iterator[LinterResult]:
         self.new_datetime = correctly_formated_datetime
 
     def fix(self) -> Iterator[LinterResult]:
+        if self.context.nasl_file.suffix == ".inc":
+            return
+
```
bjoernricks, line 104: This code is not necessary. `self.new_file_content` is never set in this case.

bjoernricks (approved):

troubadix/plugins/last_modification.py
cfi-gb: Thanks, wasn't aware of this detail. I have removed it again with the last push.

PR #309, opened 2022-07-06
Improve plugin architecture

bjoernricks:
**What**:

Drop term *pre-run`

We have two kinds of plugins:
1. FilesPlugin instances that run checks on several files at once
2. FilePlugin instance that run checks on a single file

This PR improves the plugin architecture two differentiate only between these two types where it is necessary. In general the code having to differentiate these two plugin types has been reduces significantly.

**Why**:

More generic plugin handling.

**How**:

**Checklist**:

- [x] Tests
- [ ] Documentation


troubadix/plugins/__init__.py
```diff
@@ -168,32 +171,39 @@ def __init__(
         self,
         excluded_plugins: List[str] = None,
         included_plugins: List[str] = None,
-    ) -> None:
-        super().__init__(_PLUGINS, _PRE_RUN_PLUGINS)
-
+    ):
```
y0urself, line 174:
```suggestion
    ) -> None:
```

y0urself:
I want to test this before I approve it, will take a few minutes!

y0urself (dismissed):

PR #310, opened 2022-07-07
Add: plugin script_tags_mandatory

pascalholthaus:
**What**:
Add a new plugin to check mandatory script tags.

**Why**:

**How**:

**Checklist**:

- [x] Tests
- [ ] Documentation


mbrinkhoff:
I'd argue that this is a duplicate of `script_calls_mandatory.py`, with the difference that this checks for "normal" script tags and `troubadix/plugins/script_calls_mandatory.py` checks for special script tags.

Wouldn't it be more coherent to have one plugin checking both ? At least with the current naming scheme

pascalholthaus:
What is the preferred name for the plugin?

mbrinkhoff:
I'd prefer `script_tags_mandatory`.

mbrinkhoff (approved):

PR #311, opened 2022-07-07
Add: Standalone plugin version updated

mbrinkhoff:
**What**:

Add a standalone plugin to check for commits that altered *.nasl files without updating the tags `script_version` and `last_modification`.

**Why**:

The check existed in the old QA and needs to be preserved

Closes:
- VTD-1623
- VTD-409

Possibly could also solve:
- VTD-433

**How**:

**Checklist**:

- [x] Tests
- [ ] Documentation


troubadix/standalone_plugins/version_updated.py
```diff
@@ -0,0 +1,157 @@
...
+
+import os
+import re
+import subprocess
+import sys
+from argparse import ArgumentParser, Namespace
+from pathlib import Path
+from typing import Iterable
+
+SCRIPT_VERSION_PATTERN = re.compile(
```
cfi-gb, line 31: Would it make sense to move these two pattern to helper/patterns.py or is there even some code included there for these two tags?

Moving could also help if there is some similarity with the pattern used in #312 which seems to be also about both tags.

cfi-gb: #314 is adding / has added such pattern to the `helper/patterns.py` which looks quite similar so might indeed worth to look if these could be used.

mbrinkhoff: I will look into it

mbrinkhoff: Done

tests/standalone_plugins/test_version_updated.py
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+#  test_changed_oid.py
+#
```
cfi-gb, line 5:
```suggestion
```

See the recent discussion in #327

troubadix/standalone_plugins/version_updated.py
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+#  check_last_modification.py
+#
```
cfi-gb, line 5:
```suggestion
```

See recent discussion in #327

cfi-gb (dismissed):

troubadix/standalone_plugins/version_updated.py
```diff
@@ -0,0 +1,153 @@
...
+    # git diff output uses raw bytes
+    return subprocess.run(
+        ["git"] + list(args),
+        capture_output=True,
+        encoding="latin-1",
+        check=True,
+    ).stdout
+
+
+def check_version_updated(args: Namespace) -> bool:
```
bjoernricks, line 87: Could you change this function to expect the used arguments instead of passing a Namespace object?
```suggestion
def check_version_updated(files: List[Path], commit_range: str) -> bool:
```

bjoernricks (changes requested):

troubadix/standalone_plugins/version_updated.py
bjoernricks: It is better to split parsing CLI arguments (Namespace) and the business logic (check_version_update).

mbrinkhoff: Coming

troubadix/standalone_plugins/version_updated.py
```diff
@@ -0,0 +1,154 @@
...
+def check_version_updated(files: List[Path], commit_range: str) -> bool:
+    """The script checks (via git diff) if the passed VT has changed the
+    the following tags:
+
+    - script_version("[...]");
+    - script_tag(name:"last_modification", value:"[...]");
+    """
+
+    if not files:
+        files += [
```
bjoernricks, line 96: Just to not crash if files is None (or an empty string).
```suggestion
        files = [
```

bjoernricks (approved):

PR #313, opened 2022-07-07
Fix: Script dependencies allow multiline value

mbrinkhoff:
**What**:

Fixed `script_calls_recommended.py` so that `script_dependencies` can recognize multiline values.

**Why**:

We had false-positives with the plugin.
Jira: VTD-1626

**How**:

https://stackoverflow.com/questions/33312175/matching-any-character-including-newlines-in-a-python-regex-subexpression-not-g

**Checklist**:

- [x] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/script_calls_recommended.py
+++ b/troubadix/plugins/script_calls_recommended.py
@@ -79,7 +79,7 @@ class CheckScriptCallsRecommended(FileContentPlugin):
             )
         for call in recommended_single_call:
             if not _get_special_script_tag_pattern(
-                name=call, value=".*"
+                name=call, value="(?s:.)*"
             ).search(file_content):
                 yield LinterWarning(
                     "VT does not contain the following recommended call: "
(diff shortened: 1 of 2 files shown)


troubadix/plugins/script_calls_recommended.py
```diff
@@ -79,7 +79,7 @@ def check_content(
             )
         for call in recommended_single_call:
             if not _get_special_script_tag_pattern(
-                name=call, value=".*"
+                name=call, value="(?s:.)*"
```
n-thumann, line 82:
```suggestion
                name=call, value=".*", flags=re.DOTALL
```
Is there any disadvantage in using `re.DOTALL` flag instead of this? If not, I'd prefer this, because we don't have the constrains the author on Stackoverflow has and it's easier to ready IMHO :)

n-thumann (changes requested):

troubadix/plugins/script_calls_recommended.py
mbrinkhoff: I found it nicer with the local group because I'd does make it "independent" of the outer, non-visible pattern.

Currently there is no downside to use `flags=re.DOTALL`

bjoernricks (approved):

bjoernricks (approved):

PR #327, opened 2022-07-21
Change: Add executable bit and shebang to last_modification.py

cfi-gb:
**What**:

The plugin is a "standalone" one so it can be running just like this directly from the folder and could have the same executable bit and shebang like the existing `changed_oid.py` to stay consistent.

**Why**:

Stay consistent with the other existing `changed_oid.py` standalone plugin

**How**:

N/A

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/standalone_plugins/last_modification.py
+++ b/troubadix/standalone_plugins/last_modification.py
@@ -1,3 +1,6 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
 # Copyright (C) 2022 Greenbone Networks GmbH
 #
 # SPDX-License-Identifier: GPL-3.0-or-later


bjoernricks:
No you should not do that. Please always use `poetry run troubadix-last-modification` to ensure you are running within the correct environment.

cfi-gb:
Good to know. Then i guess the existing `changed_oid.py` would need to be fixed instead to avoid confusion?

cfi-gb:
And probably troubadix/standalone_plugins/version_updated.py of #311 as well if the shebang shouldn't be included as well?

bjoernricks:
yes. running python scripts directly should be avoided. 

And just for your info, the encoding line isn't necessary at all anymore since Python 3 where utf-8 is always used. It is only necessary for Python 2.

cfi-gb:
:+1: 

Added review comments to #311 and fixed the changed_oid ones with #328

bjoernricks:
Oh and also the shebang line is/was wrong. It needs to be `/usr/bin/env python3`

cfi-gb:
Yes, indeed. But this was taken from the existing changed_oid file to stay consistent to what have been already used in the past for other standalone plugins.

I cross-checked the whole source folder and it seems the changed_oid files were the only ones which had used that wrong shebang so removing them in #328 should hopefully solve that these are getting introduced in the future again.

PR #334, opened 2022-07-28
Switch spelling plugin to files plugin

mbrinkhoff:
**What**:

Switched spelling plugin to be `FilesPlugin` based because the `FilePlugin` based approach was too slow

Closes: Jira VTD-1656

**Why**:

`FilePlugin` approach took upwards of 2 hours to finish a full run.
Current runtime (local) is about 4 minutes

**How**:

Tests have been adjusted accordingly.

**Checklist**:

- [x] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/spelling.py
+++ b/troubadix/plugins/spelling.py
@@ -55,13 +55,8 @@ class CheckSpelling(FilesPlugin):
         else:
             codespell_ignore = f"{codespell_config_path}/codespell.ignore"
 
-        # Run codespell as internal process
         batch_size = 10_000
         for i in range(0, len(self.context.nasl_files), batch_size):
-            # files_parameter = [
-            #     str(nasl_file)
-            #     for nasl_file in self.context.nasl_files[i : i + batch_size]
-            # ]
 
             files_parameters = [
                 str(nasl_file)
--- a/troubadix/plugins/spelling.py
+++ b/troubadix/plugins/spelling.py
@@ -33,16 +33,12 @@ class CheckSpelling(FilesPlugin):
     name = "check_spelling"
 
     def run(self) -> Iterator[LinterResult]:
-        """'codespell' is required to execute this step!
-        This script opens a shell in a subprocess and executes 'codespell' to
-        check the VT for spelling mistakes. An error will be thrown if
-        'codespell' is not installed or corrections could be found via
-        'codespell'.
+        """This script checks, via the codespell library, wether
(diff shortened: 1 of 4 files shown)


troubadix/plugins/spelling.py
```diff
@@ -63,177 +60,193 @@ def run(self) -> Iterator[LinterResult]:
...
-                    "--disable-colors",
-                    f"{str(self.context.nasl_file)}",
-                )
+        batch_size = 10_000
+        for i in range(0, len(self.context.nasl_files), batch_size):
+            # files_parameter = [
+            #     str(nasl_file)
+            #     for nasl_file in self.context.nasl_files[i : i + batch_size]
+            # ]
+
```
bjoernricks, line 69: Always remove commented code if it is not documentation. It just confuses other devs. If you need some old code you can always look it up in git.
```suggestion
```

troubadix/plugins/spelling.py
```diff
@@ -63,177 +60,193 @@ def run(self) -> Iterator[LinterResult]:
...
-                    if re.search(r"nD\s+==>\s+and, 2nd", line) or re.search(
-                        r"oD\s+==>\s+of", line
+            codespell_arguments = (
+                "--hard-encoding-detection",
+                "--dictionary=-",
+                f"--dictionary={codespell_additions}",
+                f"--exclude-file={codespell_exclude}",
+                f"--ignore-words={codespell_ignore}",
+                "--disable-colors",
+            ) + files_parameters
```
bjoernricks, line 83: IMHO you should use lists for concatenating the arguments

mbrinkhoff: And then just go for tuple creation right when required ?

troubadix/plugins/spelling.py
mbrinkhoff: True, forgot that

troubadix/plugins/spelling.py
bjoernricks: I suppose you did use a tuple because you tough this is required for the star operator. It's not :-)
```python
>>> a = ["a", "b"]
>>> def foo(*args):
...     print(args)
... 
>>> foo(*a)
('a', 'b')
```

mbrinkhoff: You'd be right about that

bjoernricks (approved):

PR #336, opened 2022-08-01
Fix version_updated wrong exit code

mbrinkhoff:
**What**:

Inverted the if statement in main to produce the expected exit code.

**Why**:

The check was inverted and would return 2 for success and 0 for error, this would break the QA workflow.

**How**:

Added tests to prevent regression

**Checklist**:

- [x] Tests
- [ ] Documentation


n-thumann (approved):
LGTM

PR #339, opened 2022-08-02
Fix valid_script_tag_names.py and add test cases.

cfi-gb:
**What**:

The plugin had missed to report a wrongly used script_tag() if it contained a newline because of an insufficient regex used in the custom written functions which got replaced by a helper function now.

- Closes: VTD-1667
- Part of: VTD-1616

**Why**:

The previous used functions in that plugin didn't took tags like e.g. the following below including newlines into account which caused false negatives in comparison to the old QA.

```
  script_tag(name:"vuldetect", value:"Mylongdescriptionwitha
  newline in between.");
```

**How**:

- :robot:
- Also ran `troubadix -d nasl/common/ --include-tests check_valid_script_tag_names -v` to check if no false positives are showing up

**Checklist**:

- [x] Tests
- [ ] Documentation

### DIFF
--- a/tests/plugins/test_valid_script_tag_names.py
+++ b/tests/plugins/test_valid_script_tag_names.py
@@ -1,0 +1,106 @@
+#  Copyright (c) 2022 Greenbone Networks GmbH
+#
+#  SPDX-License-Identifier: GPL-3.0-or-later
+#
+#  This program is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


cfi-gb:
Example output if a wrong (manually "broken") tag was found:

```
ℹ Start linting 97116 files ... 
ℹ Checking common/2013/gb_pcoweb_default_root_password.nasl (43591/97116)
ℹ     Results for plugin check_valid_script_tag_names
×         The script_tag name 'vulndetect' is not allowed.
ℹ Time elapsed: 0:00:24.164337
× check_valid_script_tag_names                            1        0
ℹ sum                                                     1        0
```

cfi-gb:
@greenbone/troubadix-maintainers this is ready for a review.

Notes:
- I'm not 100% sure if the unit tests are correct (mainly copy'n'pasted it from existing ones) but at least they are not failing
- The usage of `_get_tag_pattern` got taken from the existing `troubadix/plugins/script_tag_whitespaces.py`, hope that's the correct way of doing this currently

PR #350, opened 2022-08-08
Add: VT Permissions Plugin

mbrinkhoff:
**What**:

Adds a new plugin to check if VT files have the correct permissions (= not executable)

- Closes: Jira VTD-1680
- Part of: Jira VTD-1597

**Why**:

Long standing feature request that came up recently

**How**:

**Checklist**:

- [x] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/vt_file_permission.py
+++ b/troubadix/plugins/vt_file_permission.py
@@ -1,0 +1,39 @@
+# Copyright (C) 2021 Greenbone Networks GmbH
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from typing import Iterator
+
+from troubadix.plugin import FilePlugin, LinterError, LinterResult
+
+
+class CheckVTFilePermissions(FilePlugin):
+    """This script checks wether the nasl file
+    has the correct file permissions
+    """
+
+    name = "check_vt_file_permissions"
+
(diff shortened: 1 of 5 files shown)


troubadix/plugins/vt_file_permission.py
```diff
@@ -0,0 +1,39 @@
...
+    has the correct file permissions
+    """
+
+    name = "check_vt_file_permissions"
+
+    def run(self) -> Iterator[LinterResult]:
+
+        permissions = oct(self.context.nasl_file.stat().st_mode)[-3:]
+
+        if permissions != "664":
```
n-thumann, line 34:
```suggestion
        if permissions != "644":
```
See ticket and git support :)

n-thumann (changes requested):

troubadix/plugins/vt_file_permission.py
bjoernricks: The user should get a message which permissions are expected and why here too

mbrinkhoff: Nasl scripts are just expected to have `644` / are required not to be executable.

Original correction: Solved.

troubadix/plugins/vt_file_permission.py
```diff
@@ -0,0 +1,41 @@
...
+    has the correct file permissions
+    """
+
+    name = "check_vt_file_permissions"
+
+    def run(self) -> Iterator[LinterResult]:
+
+        permissions = oct(self.context.nasl_file.stat().st_mode)[-3:]
+
+        if permissions != "644":
```
n-thumann, line 34: Mh, maybe it makes sense to only check for the executable bit (since Git only keeps track of it and doesn't care about the others).
Also, Git seems to be creating the files with `664` by default, so I have a ton of files that don't comply with this check anyways. Because of that the unittest is failing for me, because Git created the `ok_permissions.nasl` file with `664`.
Don't we need to take this into account in both the ticket and the corresponding Confluence page, @cfi-gb?

cfi-gb: To the best of my knowledge all modern / standard Linux systems are using a default `umask` of `0022`:

```
$ umask
0022
```

which translates to the file permission of `644` which are checked here. Not sure why your system behaves differently but i can't find any files in the recent VTs repository having permissions of `664`.

mbrinkhoff: Default umask on my Ubuntu 22.04: 
```
➜  ~ umask   
002
```

mbrinkhoff: Implemented

n-thumann (approved):

PR #359, opened 2022-08-16
Add Alma Linux script family and OID

mbrinkhoff:
**What**:

Adds the script family for Alma Linux and the OID pattern used by Notus-Generator

**Why**:

Part of: VTD-1310

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/script_family.py
+++ b/troubadix/plugins/script_family.py
@@ -26,6 +26,7 @@ from troubadix.plugin import FileContentPlugin, LinterError, LinterResult
 
 VALID_FAMILIES = [
     "AIX Local Security Checks",
+    "Alma Linux Local Security Checks",
     "Amazon Linux Local Security Checks",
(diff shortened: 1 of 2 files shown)


troubadix/plugins/script_family.py
```diff
@@ -87,6 +87,7 @@
     "Web application abuses",
     "Windows",
     "Windows : Microsoft Bulletins",
+    "Alma Linux Local Security Checks",
```
n-thumann, line 90: Can you insert it, so that it's sorted correctly? You did the same thing for Rocky as well 👀

n-thumann (changes requested):

n-thumann (approved):

troubadix/plugins/script_family.py
mbrinkhoff: Fixed

PR #369, opened 2022-08-22
Reworked illegal characters check

mbrinkhoff:
**What**:

Reworked the illegal characters check so that it distinguishes between forbidden characters and characters that only should create a warning

**Why**:

The previous behavior from #53 only created warnings, even if the file content was illegal.

**How**:

**Checklist**:

- [x] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/illegal_characters.py
+++ b/troubadix/plugins/illegal_characters.py
@@ -16,11 +16,17 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import re
-from typing import Iterator, Union
+from typing import Iterator, List
 
 from troubadix.helper import CURRENT_ENCODING
 from troubadix.helper.patterns import get_common_tag_patterns
-from troubadix.plugin import FilePlugin, LinterError, LinterFix, LinterResult
+from troubadix.plugin import (
+    FilePlugin,
+    LinterError,
+    LinterFix,
+    LinterResult,
+    LinterWarning,
+)
 
 # import magic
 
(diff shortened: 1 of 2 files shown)


n-thumann (changes requested):

troubadix/plugins/illegal_characters.py
```diff
@@ -30,27 +36,60 @@
 # | =               are delimiter in the internal VT cache
 #                   (e.g. Tag1=Foo|Tag2=Bar|Tag3=Baz)
 #                   in all script_tag(name:"", value:"")
-FORBIDDEN_CHARS = ["|", "=", ";"]
+FORBIDDEN_CHARS = ["|", ";"]
+WARNING_CHARS = ["="]
 
 # replacement character will be a whitespace
 REPLACE_CHAR = " "
```
cfi-gb, line 43: Not sure if this would be in the scope of this PR (that's my only as a general comment and not a change request):

Replacing any of these chars with just a space when passing `--fix` to Troubadix should be avoided because it changes the meaning of the metadata (e.g. a `|` is often used in code examples of LSCs).

Would it be possible "while on it" to have some kind of array in `FORBIDDEN_CHARS` and `WARNING_CHARS` which includes the char which is an error or warning and the char which should be used for a replacement (e.g. replacing `|` with `<pipe>`?

n-thumann: Sounds good to me, so sth. like
```
FORBIDDEN_CHARS = {
    "|": "<pipe>",
    ";": "foobar"
}
```
Then the check for existence of a character could be kept as is and only replacing it would need to be adjusted (accessing the corresponding dict index instead)

mbrinkhoff: In conjunction with Nicolas request from above, regarding a single check:
We could create a dictionary like this: 
```
forbidden_chars = {
    <char>: (<ErrorType>, <Replacement>)
}
```

Which would wrap warning and error in the same list and allow a specific replace per forbidden char.

troubadix/plugins/illegal_characters.py
```diff
@@ -30,27 +36,60 @@
...
 
 
-def check_match(match: re.Match) -> Union[str, None]:
-    """Replace all FORBIDDEN characters with the replace character defined.
+def check_forbidden(match: re.match) -> List[str]:
```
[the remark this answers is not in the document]
mbrinkhoff: See answer below Christians comment.

troubadix/plugins/illegal_characters.py
```diff
@@ -30,27 +36,60 @@
...
+    """Check the given tag for forbidden characters
+
+    Args:
+        match (re.match): The tag to check
+
+    Returns:
+        List[str]: The list of forbidden characters that were found
+    """
+    return [
+        f"'{char}'" for char in FORBIDDEN_CHARS if char in match.group("value")
```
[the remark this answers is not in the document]
mbrinkhoff: True indeed, fix coming

troubadix/plugins/illegal_characters.py
mbrinkhoff: @cfi-gb Any desired replacement for `;` ?

n-thumann: Side note: In Notus Generator we replace `;` with `,`

cfi-gb: Replacing `;` with `,` by default should work. The dev would need to review the changes anyway and could still change it if it should be replaced by e.g. `<semicolon>`.

mbrinkhoff: `,` it is then (for now)

troubadix/plugins/illegal_characters.py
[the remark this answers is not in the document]
mbrinkhoff: Fixed, also added a test for that

n-thumann (approved):

PR #388, opened 2022-09-20
Add: Additional exclusions to the spelling plugin.

cfi-gb:
**What**:
See title

Notes:
- No new release required, there will be at least one follow-up PR for syncing the current `codespell.*` files from the vulerability-tests repo.
- Parity PR to greenbone/vulnerability-tests#1319
- Some further improvements for this plugin are not tracked in VTD-1773

**Why**:
Exclude a few false positives in a more generic way.

**How**:
Run the spelling plugin against the whole feed and less spelling issues should be reported.

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/spelling.py
+++ b/troubadix/plugins/spelling.py
@@ -259,6 +259,19 @@ class CheckSpelling(FilesPlugin):
                         if re.search(r"[Uu]nsecure\s+==>\s+[Ii]nsecure", line):
                             continue
 
+                    # NAM / nam is the abbreviation of these products. In
+                    # netop_infopublic.nasl there is a "nam" parameter.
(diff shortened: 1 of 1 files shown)


mbrinkhoff:
@cfi-gb Should we transform this into a list too ? Looks like a candidate to get a bit straightened too

cfi-gb:
> @cfi-gb Should we transform this into a list too ? Looks like a candidate to get a bit straightened too

This would be indeed an improvement possibility for the future (i would leave that out of this PR for now), e.g. `_check_false_positive` from the `grammar.py` could be moved into the `troubadix/helper/helper.py` file and used for both plugins.

PR #394, opened 2022-09-28
Fix: Allow multiline dependencies with inline comments

mbrinkhoff:
**What**:

Fixes false positives in the dependency check if inline comments are present in the parameter list

**Why**:

Closes: VTD-1782

**How**:

Output before PR:
```
ℹ Start linting 131121 files ... 
ℹ Time elapsed: 0:00:10.588791
× check_dependencies                                      2      433
ℹ sum                                                     2      433
```

Output with PR:
```
ℹ Start linting 131121 files ... 
ℹ Time elapsed: 0:00:10.866805
⚠ check_dependencies                                      0      433
ℹ sum                                                     0      433
```

**Checklist**:

- [x] Tests
- [ ] Documentation

### DIFF
--- a/tests/plugins/test_dependencies.py
+++ b/tests/plugins/test_dependencies.py
@@ -196,16 +196,16 @@ class CheckDependenciesTestCase(PluginTestCase):
     def test_inline_comment_dependency(self):
         with self.create_directory() as tmpdir:
             path = tmpdir / "file.nasl"
-            example = tmpdir / "common" / "example.inc"
-            example2 = tmpdir / "common" / "example2.inc"
+            example = tmpdir / "common" / "example.nasl"
+            example2 = tmpdir / "common" / "example2.nasl"
             example.parent.mkdir(parents=True)
             example.touch()
             example2.touch()
             content = (
                 'script_tag(name:"cvss_base", value:"4.0");\n'
                 'script_tag(name:"summary", value:"Foo Bar...");\n'
-                'script_dependencies("example.inc", #Comment\n'
-                '"example2.inc");\n'
+                'script_dependencies("example.nasl", #Comment\n'
+                '"example2.nasl");\n'
             )
             fake_context = self.create_file_plugin_context(
                 nasl_file=path, file_content=content, root=tmpdir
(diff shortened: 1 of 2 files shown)


cfi-gb:
Nice, thanks a lot for taking care :+1: 

> runtime has increased quite a lot

At least in the given output i don't see a huge increase of runtime

tests/plugins/test_dependencies.py
```diff
@@ -192,3 +192,47 @@ def test_dependency_missing_newline(self):
                 "not be found within the VTs.",
                 results[0].message,
             )
+
+    def test_inline_comment_dependency(self):
+        with self.create_directory() as tmpdir:
+            path = tmpdir / "file.nasl"
+            example = tmpdir / "common" / "example.inc"
```
cfi-gb, line 199: Better use the `.nasl` file ending here and in the other test as `script_dependencies()` doesn't use `.inc` files.

cfi-gb (dismissed):

mbrinkhoff:
> Nice, thanks a lot for taking care +1
> 
> > runtime has increased quite a lot
> 
> At least in the given output i don't see a huge increase of runtime

Removed that comment, I first ran the it with a debugger attached and ended up with a > 1 minute runtime.

tests/plugins/test_dependencies.py
mbrinkhoff: Done

troubadix/plugins/dependencies.py
```diff
@@ -58,9 +58,13 @@ def run(
...
                 # Remove single and/or double quotes, spaces
                 # and create a list by using the comma as a separator
-                dependencies = re.sub(
-                    r'[\'"\s]', "", match.group("value")
-                ).split(",")
+                # additionally, check and filter for inline comments
+                dependencies = []
+
+                for line in match.group("value").splitlines():
+                    subject = line[: line.find("#")] if "#" in line else line
```
n-thumann, line 65:
```suggestion
                    subject = line[: line.index("#")] if "#" in line else line
```
nitpick: Can we use `index` instead? I think it clarifies a bit more, what it done here and we're checking for the existence of `#` anyways.

mbrinkhoff: Yes

n-thumann (dismissed):

pbrozmann (approved):

PR #420, opened 2022-11-09
Reduce no_solution noise

mbrinkhoff:
**What**:

Removes the summary output from the `no_solution` check.

**Why**:

Part of: VTD-1828

**How**:

**Checklist**:

- [x] Tests
- [ ] Documentation


cfi-gb:
IMHO the whole plugin could be completely dropped in favor of the new standalone one #421.

PR #432, opened 2022-11-23
Add: Check for malformed script_dependency  values

mbrinkhoff:
**What**:

Adds a plugin to check for dependencies which are malformed by being enumerated within a single pair of quotes.

**Why**:

Closes: VTD-444

**How**:

**Checklist**:

- [x] Tests
- [ ] Documentation

### DIFF
--- a/tests/plugins/test_malformed_dependencies.py
+++ b/tests/plugins/test_malformed_dependencies.py
@@ -31,7 +31,7 @@ class CheckMalformedDependenciesTestCase(PluginTestCase):
         content = (
             'script_tag(name:"cvss_base", value:"4.0");\n'
             'script_tag(name:"summary", value:"Foo Bar.");\n'
-            'script_dependencies("example.inc", "example2.inc");\n'
+            'script_dependencies("example.nasl", "example2.nasl");\n'
         )
         fake_context = self.create_file_plugin_context(
             nasl_file=path, file_content=content, root=here
@@ -47,7 +47,7 @@ class CheckMalformedDependenciesTestCase(PluginTestCase):
         content = (
             'script_tag(name:"cvss_base", value:"4.0");\n'
             'script_tag(name:"summary", value:"Foo Bar.");\n'
-            'script_dependencies("example.inc,example2.inc");\n'
+            'script_dependencies("example.nasl,example2.nasl");\n'
         )
         fake_context = self.create_file_plugin_context(
             nasl_file=path, file_content=content, root=here
@@ -58,7 +58,7 @@ class CheckMalformedDependenciesTestCase(PluginTestCase):
 
         expected_result = (
             "The script dependency value is malformed and contains a "
-            "comma in the dependency value: 'example.inc,example2.inc'"
+            "comma in the dependency value: 'example.nasl,example2.nasl'"
         )
 
         self.assertEqual(len(results), 1)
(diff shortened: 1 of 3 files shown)


troubadix/plugins/malformed_dependencies.py
```diff
@@ -0,0 +1,68 @@
...
+        dependencies_pattern = _get_special_script_tag_pattern(
+            "dependencies", flags=re.DOTALL | re.MULTILINE
+        )
+
+        file_content = self.context.file_content
+
+        matches = dependencies_pattern.finditer(file_content)
+
+        for match in matches:
+            if match:
```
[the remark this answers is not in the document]
mbrinkhoff: Good point actually, at least in this plugin we can start doing better.

troubadix/plugins/malformed_dependencies.py
```diff
@@ -0,0 +1,68 @@
...
+
+                tag_value = (
+                    f'"{match.group("value")}"' if match.group("value") else ""
+                )
+
+                quote_count = len(
+                    [char for char in tag_value if char in ['"', "'"]]
+                )
+                comma_count = len([char for char in tag_value if char == ","])
```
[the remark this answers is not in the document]
mbrinkhoff: Done

troubadix/plugins/malformed_dependencies.py
```diff
@@ -0,0 +1,68 @@
...
+
+                required_ratio = quote_count / 2 - 1
+                if quote_count >= 2 and comma_count != required_ratio:
```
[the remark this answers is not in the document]
mbrinkhoff: Done

troubadix/plugins/malformed_dependencies.py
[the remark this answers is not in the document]
mbrinkhoff: Done

troubadix/plugins/malformed_dependencies.py
```diff
@@ -0,0 +1,87 @@
...
+                if "," in dependency_value:
+                    yield LinterError(
+                        "The script dependency value is malformed and "
+                        "contains a comma in the dependency value: "
+                        f"'{dependency_value}'",
+                        file=self.context.nasl_file,
+                        plugin=self.name,
+                    )
+
+                if not INNER_WHITESPACE_PATTERN.match(dependency_value):
```
n-thumann, line 80: What do we need a RegEx for and cannot use `if " " in dependency_value`? Doesn't `dependency_value` always contain another NASL filename, so no spaces at all?

mbrinkhoff: Just going from the assumption that leading / trailing whitespace is at least tolerated by openvas.

n-thumann: Just checked with `openvas-nasl-lint` and it doesn't complain about it either, so it's probably fine.

n-thumann (dismissed):

troubadix/plugins/malformed_dependencies.py
cfi-gb: Note that i don't think that `openvas-nasl-lint` is checking the dependencies at all, that's why we have the related Troubadix checks in place.

mbrinkhoff: We checked and there are currently no trailing / leading / any whitespaces in the feed dependency values.

Test now checks for any whitespace and throws an error if one is found.

tests/plugins/test_malformed_dependencies.py
```diff
@@ -0,0 +1,168 @@
...
+here = Path(__file__).parent
+
+
+class CheckMalformedDependenciesTestCase(PluginTestCase):
+    def test_ok(self):
+        path = here / "file.nasl"
+        content = (
+            'script_tag(name:"cvss_base", value:"4.0");\n'
+            'script_tag(name:"summary", value:"Foo Bar.");\n'
+            'script_dependencies("example.inc", "example2.inc");\n'
```
cfi-gb, line 34:
```suggestion
            'script_dependencies("example.nasl", "example2.nasl");\n'
```

For the sake of completeness i would change all examples to `.nasl` because `.inc` can't be dependencies at all.

mbrinkhoff: Done

n-thumann (approved):

PR #444, opened 2022-12-08
Fix: Roman encoding problem

mbrinkhoff:
**What**:

Fixes problems with detecting wrong encodings on attic files.

**Why**:

Fixes: VTD-1854

**How**:

**Checklist**:

- [x] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/plugins/encoding.py
+++ b/troubadix/plugins/encoding.py
@@ -38,21 +38,21 @@ class CheckEncoding(LineContentPlugin):
         lines: Iterable[str],
     ) -> Iterator[LinterResult]:
 
-        for detection in chardet.detect_all(nasl_file.read_bytes()):
-            encoding = detection.get("encoding")
-            language = detection.get("language")
-            if (
-                encoding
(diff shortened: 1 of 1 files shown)


troubadix/plugins/encoding.py
```diff
@@ -38,21 +38,21 @@ def check_lines(
...
-                    f"encoding is {encoding}.",
-                    file=nasl_file,
-                    plugin=self.name,
-                )
-                break
+        content = nasl_file.read_bytes()
+
+        detection = chardet.detect(content)
+        encoding = detection.get("encoding")
+        if encoding and encoding not in [
```
n-thumann, line 45:
```suggestion
        if encoding not in [
```
nit: Any reason to keep the check for `encoding`?

mbrinkhoff: It is, for some reason, empty sometimes and thus the empty encoding is not in the array, triggering a FP

n-thumann (approved):

PR #465, opened 2023-01-11
Change: Add exit code to determinate wether VTs to report were found

mbrinkhoff:
**What**:

Adds an exit code that returns wether VTs have been found to report (0 = none, 1 = found)

**Why**:

Part of: VTD-1871

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/standalone_plugins/no_solution.py
+++ b/troubadix/standalone_plugins/no_solution.py
@@ -279,7 +279,7 @@ def main():
 
     print_report(term, summary, arguments.threshold, root, found_vts)
 
-    return 1 if len(found_vts) > 0 else 0
+    sys.exit(1 if found_vts > 0 else 0)
 
 
 if __name__ == "__main__":
--- a/troubadix/standalone_plugins/no_solution.py
+++ b/troubadix/standalone_plugins/no_solution.py
@@ -227,9 +227,8 @@ def print_report(
     summary: Iterable[Tuple[int, List[Tuple[Path, str, datetime, datetime]]]],
     threshold: int,
     root: Path,
+    total: int,
 ):
-    total = sum(len(vts) for _, vts in summary)
-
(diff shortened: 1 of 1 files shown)


troubadix/standalone_plugins/no_solution.py
```diff
@@ -275,7 +275,11 @@ def main():
 
     summary = check_no_solutions(files, milestones, arguments.snooze)
 
-    print_report(term, summary, arguments.threshold, root)
+    found_vts = sum(len(entries) for _, entries in summary)
+
+    print_report(term, summary, arguments.threshold, root, found_vts)
+
+    return 1 if len(found_vts) > 0 else 0
```
n-thumann, line 282:
```suggestion
    sys.exit(1 if found_vts > 0 else 0)
```
Can't take `len` of an int and the return code is ignore for me, because `main()` is called directly by poetry (and AFAIK also without it when installed as usual)

troubadix/standalone_plugins/no_solution.py
```diff
@@ -275,7 +275,11 @@ def main():
...
 
 
 if __name__ == "__main__":
```
n-thumann, line 285: Maybe remove this one then?

n-thumann (changes requested):

troubadix/standalone_plugins/no_solution.py
mbrinkhoff: Yeah, VS code didnt work like that apparently. Fixed

troubadix/standalone_plugins/no_solution.py
mbrinkhoff: This breaks the VS code debugger on my side

n-thumann: Ah, let's keep it then

PR #470, opened 2023-01-12
Fix crashing behavior

mbrinkhoff:
**What**:

`troubadix-no-solution` now catches internal errors and just displays the error message

**Why**:

Closes: VTD-1896

**How**:

**Checklist**:

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/troubadix/standalone_plugins/no_solution.py
+++ b/troubadix/standalone_plugins/no_solution.py
@@ -289,7 +289,7 @@ def main():
 
         print(f"troubadix-no-solution encountered an error: {e}")
 
-        sys.exit(-1)
+        sys.exit(2)
 
 
 if __name__ == "__main__":
--- a/troubadix/standalone_plugins/no_solution.py
+++ b/troubadix/standalone_plugins/no_solution.py
@@ -282,4 +282,4 @@ def main():
 
 
 if __name__ == "__main__":
-    sys.exit(main())
+    main()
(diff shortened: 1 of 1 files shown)


troubadix/standalone_plugins/no_solution.py
```diff
@@ -261,26 +261,37 @@ def print_report(
...
+        print_report(term, summary, arguments.threshold, root, found_vts)
+
+        sys.exit(1 if found_vts > 0 else 0)
+
+    # pylint: disable=broad-except
+    except Exception as e:
+
+        print(f"troubadix-no-solution encountered an error: {e}")
+
+        sys.exit(-1)
```
n-thumann, line 293:
```suggestion
        sys.exit(1)
```
Can we use something different here? Because of the exit code being an unsigned integer, it'll be handled as `255` anyways:
```
(.venv) ➜  troubadix git:(no_solution_fix_dir_crash) ✗ poetry run troubadix-no-solution -d ../vulnerability-tests/nasl -m 99
troubadix-no-solution encountered an error: Test!
(.venv) ➜  troubadix git:(no_solution_fix_dir_crash) ✗ echo $?
255
```
Since we only distinguish between zero and non-zero in our pipeline, how about just using `1`?

n-thumann (changes requested):

troubadix/standalone_plugins/no_solution.py
mbrinkhoff: Going with `2` since `1` is already occupied by "found stuff to fix"

n-thumann (approved):
Checked that `KeyboardInterrupt` / SIGINT is still possible, works as expected.

PR #492, opened 2023-02-09
Add: changed-packages standalone plugin 

n-thumann:
## What
This PR adds `changed-packages` as a standalone plugin to Troubadix.

## Why
To give NASL developers to review large PRs easier and quickly check the changed packages.

## References

VTA-397

## Checklist

- [x] Tests

### DIFF
--- a/troubadix/standalone_plugins/changed_packages/changed_packages.py
+++ b/troubadix/standalone_plugins/changed_packages/changed_packages.py
@@ -21,7 +21,7 @@ import sys
 from argparse import ArgumentParser, Namespace
 from pathlib import Path
 from subprocess import CalledProcessError
-from typing import Iterable
+from typing import Iterable, List
 
 from pontos.terminal.terminal import ConsoleTerminal
 
@@ -61,7 +61,7 @@ def compare(old_content: str, content: str):
     return missing_packages, new_packages
 
 
-def filter_reasons(packages: list[Package], reasons: Iterable[Reasons]):
+def filter_reasons(packages: List[Package], reasons: Iterable[Reasons]):
     return [
         package
         for package in packages
@@ -71,8 +71,8 @@ def filter_reasons(packages: list[Package], reasons: Iterable[Reasons]):
 
 
 def print_results(
-    missing_packages: list[Package],
-    new_packages: list[Package],
+    missing_packages: List[Package],
+    new_packages: List[Package],
     file: Path,
     terminal: ConsoleTerminal,
 ):
@@ -88,7 +88,7 @@ def print_results(
 
 
 def print_packages(
-    packages: list[Package],
+    packages: List[Package],
     terminal: ConsoleTerminal,
 ):
     with terminal.indent():
--- a/troubadix/standalone_plugins/changed_packages/changed_packages.py
+++ b/troubadix/standalone_plugins/changed_packages/changed_packages.py
@@ -127,11 +127,7 @@ def parse_args() -> Namespace:
         type=file_type,
         default=[],
         required=True,
-        help=(
-            "List of files to diff."
-            "If empty use all files added or modifyed in the"
-            " commit range"
-        ),
+        help="List of files to check.",
     )
     parser.add_argument(
         "--start-commit",
--- a/troubadix/standalone_plugins/changed_packages/changed_packages.py
+++ b/troubadix/standalone_plugins/changed_packages/changed_packages.py
@@ -1,0 +1,205 @@
+# Copyright (C) 2023 Greenbone Networks GmbH
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import re
+import sys
+from argparse import ArgumentParser, Namespace
+from pathlib import Path
+from subprocess import CalledProcessError
+from typing import Iterable
+
+from pontos.terminal.terminal import ConsoleTerminal
+
+from troubadix.argparser import file_type
+from troubadix.standalone_plugins.changed_packages.marker import (
+    AddedEpoch,
+    AddedRelease,
+    AddedUdeb,
+    ChangedUpdate,
+    DroppedArchitecture,
+)
+from troubadix.standalone_plugins.changed_packages.package import (
+    Package,
+    Reasons,
+)
+from troubadix.standalone_plugins.common import git
+
+PACKAGE_CHECK_PATTERN = re.compile(
+    r'isdpkgvuln\(pkg:"(?P<package>[^"]+)", ver:"(?P<version>[^"]+)", '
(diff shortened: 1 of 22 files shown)


troubadix/standalone_plugins/changed_packages/changed_packages.py
```diff
@@ -0,0 +1,205 @@
...
+    parser.add_argument(
+        "--files",
+        nargs="+",
+        type=file_type,
+        default=[],
+        required=True,
+        help=(
+            "List of files to diff."
+            "If empty use all files added or modifyed in the"
+            " commit range"
```
cfi-gb, line 133:
```suggestion
            "List of files to diff. "
            "If empty use all files added or modified in the commit range."
```

Same fix as in #490

n-thumann: Fixed by adjusting the help text.

troubadix/standalone_plugins/changed_packages/package.py
```diff
@@ -0,0 +1,91 @@
...
+    def __lt__(self, other: "Package") -> bool:
+        # Sort by release first, then the other fields
+        if self.release != other.release:
+            return self.release < other.release
+        if self.name != other.name:
+            return self.name < other.name
+        if self.version != other.version:
+            return self.version < other.version
+
+        return False
```
mbrinkhoff, line 77: Maybe I am trying to make a simple thing hard:

`return self.release < other.release or self.name < other.name or self.version < other.version`

troubadix/standalone_plugins/changed_packages/package.py
```diff
@@ -0,0 +1,91 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+from argparse import ArgumentError
+from dataclasses import dataclass, field
+from enum import Enum
+from typing import Dict
```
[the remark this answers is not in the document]
n-thumann: Sadly no, because the CI requires older versions to be supported (see the commits reverting my usage of `dict`)

troubadix/standalone_plugins/changed_packages/package.py
n-thumann: The issue with this is the `or`, because this will still continue, even if `self.release > other.release`.
Theoretically the equivalent would be `self.release < other.release or (self.release == other.release and self.name < other.name) or (self.release == other.release and self.name == other.name and self.version < other.version)`, if you prefer this one :trollface:
But: Thanks for bringing this up, I added a unit test for cover this!

troubadix/standalone_plugins/changed_packages/changed_packages.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def get_packages(content: str):
+    package_checks = PACKAGE_CHECK_PATTERN.findall(content)
+    result = set(
+        [
+            Package(name=name, version=version, release=release)
+            for name, version, release in package_checks
+        ]
+    )
```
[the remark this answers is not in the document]
n-thumann: That's correct, not sure why I always miss this. Fixed.

troubadix/standalone_plugins/changed_packages/package.py
```diff
@@ -0,0 +1,91 @@
...
+    def __str__(self) -> str:
+        result = f"{self.name : <50} {self.version : <40} {self.release : <10}"
+
+        reasons = ", ".join(
+            [
+                f"{change}"
+                f"{' in new package' if direction == Direction.PASSIVE else ''}"
+                for change, direction in self.reasons.items()
+            ]
+        )
```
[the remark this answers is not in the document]
n-thumann: Fixed

troubadix/standalone_plugins/changed_packages/package.py
mbrinkhoff: Thanks for clarification!

troubadix/standalone_plugins/changed_packages/changed_packages.py
```diff
@@ -0,0 +1,201 @@
...
+
+    if len(result) != len(package_checks):
```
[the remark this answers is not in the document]
n-thumann: Fixed via improved error handling.

troubadix/standalone_plugins/changed_packages/changed_packages.py
```diff
@@ -0,0 +1,201 @@
...
+    terminal.info(
+        f"Checking {len(args.files)} file(s) from {args.start_commit} to HEAD"
+    )
+
+    for file in args.files:
+        try:
+            old_content = git("show", f"{args.start_commit}:{file}")
+        except CalledProcessError:
+            terminal.error(f"Could not find {file} at {args.start_commit}. ")
+            sys.exit(1)
```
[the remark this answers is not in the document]
n-thumann: Agree, probably doesn't make sense. Fixed via improved error handling.

troubadix/standalone_plugins/changed_packages/changed_packages.py
```diff
@@ -0,0 +1,201 @@
...
+
+def get_merge_base():
+    return git("merge-base", "main", "HEAD").strip()
+
+
+def parse_args() -> Namespace:
+    parser = ArgumentParser(
+        description="Check for changed packages in dpkg-based LSCs",
+    )
+    parser.add_argument(
```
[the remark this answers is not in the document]
n-thumann: I'd like to omit this feature, as long as the NASL devs aren't asking for it to keep the codebase rather small and you can still do it as a one-liner using e.g. `troubadix-changed-packages --files $(git diff main...HEAD --name-only | grep "nasl")`.
We don't have that argument on other tools like change-oid, no-solution and version-updated too.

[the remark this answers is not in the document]
mbrinkhoff: Yeah I will put it on my todo list

troubadix/standalone_plugins/changed_packages/changed_packages.py
```diff
@@ -168,12 +165,16 @@ def main():
...
+            missing_packages, new_packages = compare(old_content, content)
         except CalledProcessError:
-            terminal.error(f"Could not find {file} at {args.start_commit}. ")
-            sys.exit(1)
-
-        content = git("show", f"HEAD:{file}")
-        missing_packages, new_packages = compare(old_content, content)
+            terminal.error(
+                f"Could not find {file} at {args.start_commit} or HEAD"
+            )
```
mbrinkhoff, line 173:
```suggestion
            terminal.warn(
                f"Could not find {file} at {args.start_commit} or HEAD"
            )
```

Is is `warn` or `warning` ?

mbrinkhoff (changes requested):

troubadix/standalone_plugins/changed_packages/changed_packages.py
n-thumann: `ConsoleTerminal` only has `warning´. Were you thinking of https://docs.python.org/3/library/logging.html#logging.warning, which also has `warn` (although being deprecated)?

mbrinkhoff: Updated code suggestion

mbrinkhoff: I'd argue that a file not existing at a previous state is not an error condition (and at best a warning), but I am not deep enough into this, so its up to you.

mbrinkhoff (approved):

troubadix/standalone_plugins/changed_packages/changed_packages.py
n-thumann: The issue with not finding the previous state is that we cannot perform a comparison then and changes to the packages can slip through unnoticed.
I expected that not finding it only occurs, because a user chose the wrong start commit (e.g. where the NASL files have a different name).

PR #529, opened 2023-03-29
Change: security_message check to scan for security_message calls

mbrinkhoff:
## What

Changes the `security_message` plugin to check that either:

- A `severity` is present, as is a call to `security_message`
- A `severity` is absent, as is a call to `security_message`

## Why

Closes: VTD-1930

## Checklist

- [x] Tests

### DIFF
--- a/troubadix/plugins/security_messages.py
+++ b/troubadix/plugins/security_messages.py
@@ -36,7 +36,8 @@ def _file_contains_security_message(file_content: str) -> bool:
         file_content (str): The content of the VT
     """
     return any(
-        sec_msg in file_content for sec_msg in security_message_implementations
+        security_message in file_content
+        for security_message in SECURITY_MESSAGE_IMPLEMENTATIONS
     )
 
 
--- a/troubadix/plugins/security_messages.py
+++ b/troubadix/plugins/security_messages.py
@@ -20,7 +20,7 @@ from typing import Iterator
 from troubadix.helper.patterns import ScriptTag, get_script_tag_pattern
 from troubadix.plugin import FileContentPlugin, LinterError, LinterResult
 
-security_message_implementations = [
+SECURITY_MESSAGE_IMPLEMENTATIONS = [
     "security_message",
     "http_check_remote_code",
     "citrix_xenserver_check_report_is_vulnerable",
--- a/troubadix/plugins/security_messages.py
+++ b/troubadix/plugins/security_messages.py
@@ -29,6 +29,12 @@ security_message_implementations = [
 
 
 def _file_contains_security_message(file_content: str) -> bool:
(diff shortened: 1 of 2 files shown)


troubadix/plugins/security_messages.py
```diff
@@ -14,34 +14,69 @@
...
 #  You should have received a copy of the GNU General Public License
 #  along with this program.  If not, see <http://www.gnu.org/licenses/>.
-import re
 from pathlib import Path
 from typing import Iterator
 
 from troubadix.helper.patterns import ScriptTag, get_script_tag_pattern
 from troubadix.plugin import FileContentPlugin, LinterError, LinterResult
 
+security_message_implementations = [
```
n-thumann, line 23:
```suggestion
SECURITY_MESSAGE_IMPLEMENTATIONS = [
```
Because it's used as a constant

troubadix/plugins/security_messages.py
```diff
@@ -14,34 +14,69 @@
...
+def _file_contains_security_message(file_content: str) -> bool:
+    """Checks wether a VT content contains a call to security_message
+    or any function known to implement it
+
+    Args:
+        file_content (str): The content of the VT
+    """
+    return any(
+        sec_msg in file_content for sec_msg in security_message_implementations
+    )
```
n-thumann, line 40:
```suggestion
    return any(
        security_message in file_content
        for security_message in SECURITY_MESSAGE_IMPLEMENTATIONS
    )
```
I'd prefer using full variable names rather than abbreviations

n-thumann (approved):

troubadix/plugins/security_messages.py
```diff
@@ -14,34 +14,70 @@
...
-
+            nasl_file (Path): The VTs path
+            file_content (str): The content of the VT
         """
-        if nasl_file.suffix == ".inc":
+        deprecated_pattern = get_script_tag_pattern(
+            script_tag=ScriptTag.DEPRECATED
+        )
+        if deprecated_pattern.match(file_content):
             return
```
cfi-gb, line 61: It seems this *might* not work as expected (e.g. a problem in the helper function) because in e.g. greenbone/vulnerability-tests#4117 i'm seeing a false positive for a deprecated VT having this tag:

```
  script_tag(name:"deprecated", value:TRUE);
```

mbrinkhoff: On it

cfi-gb:
```suggestion
        deprecated_pattern = get_script_tag_pattern(
            script_tag=ScriptTag.DEPRECATED
        )
        if deprecated_pattern.search(file_content):
            return
```

Seems this is working, at least for me.

PR #550, opened 2023-04-20
Change: Exclude template and testing VTs from some checks.

cfi-gb:
## What
See title

To test run Troubadix with something like e.g. the following:

```
troubadix -vv --files template.nasl policy_control_template.nasl path/to/test_version_func_inc.nasl path/to/test_ipv6_packet_forgery.nasl
```

and see the false positives reports for these test / template files. Repeat it from this branch and against the VTs in the branch of the PR linked below and the false positives should be gone.

Notes:
- I have kept this a little bit "basic" on purpose because there are only three VTs / nasl files which currently needs to be excluded. If the exclusion for e.g. different keywords or similar are required we might want to handle this in a more advanced way in the future
- Using such "exclusion" keywords makes it easier to add additional testing files / templates without touching Troubadix every time
- `template_nd_test_files_fps` naming is TBD, better suggestions are welcome
- Inspired by the `# pylint: disable=something` of Pylint (see https://pylint.readthedocs.io/en/latest/user_guide/messages/message_control.html)
- The `copyright_year.py` is also a little bit special as it is a `LineContentPlugin`. I have raised VTD-1964 to make it a `FileContentPlugin` and then we could use the more generic approach similar to the others once that task has been tackled
- In one case (dependency_category_order.py)  `.inc` files didn't got excluded and i have also added a check for that
- In a few other cases e.g. `root = self.context.root` got defined but then the plugin might have returned early without using it so a few of these cases have been moved a little bit more down "while on it". Haven't explicitly mentioned that in the commit message because it's a minor change which IMHO doesn't worth it to be included in the changelog

## Why
To make PRs for the files in question (like e.g. greenbone/vulnerability-tests#4445) to pass without having an admin overwriting the (wrongly) failed checks

## References
None

### DIFF
--- a/troubadix/standalone_plugins/version_updated.py
+++ b/troubadix/standalone_plugins/version_updated.py
@@ -23,6 +23,7 @@ from argparse import ArgumentParser, Namespace
 from pathlib import Path
 from typing import Iterable, List
 
+from troubadix.helper import is_ignore_file
 from troubadix.helper.patterns import (
     LAST_MODIFICATION_ANY_VALUE_PATTERN,
     SCRIPT_VERSION_ANY_VALUE_PATTERN,
@@ -35,6 +36,12 @@ SCRIPT_VERSION_PATTERN = re.compile(
 SCRIPT_LAST_MODIFICATION_PATTERN = re.compile(
     r"^\+\s*" + LAST_MODIFICATION_ANY_VALUE_PATTERN, re.MULTILINE
 )
+
+_IGNORE_FILES = [
+    "/template.nasl",
+    "/policy_control_template.nasl",
+    "test_version_func_inc.nasl",
+]
 
 
 def file_type(string: str) -> Path:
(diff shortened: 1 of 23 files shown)


mbrinkhoff (approved):

troubadix/standalone_plugins/version_updated.py
```diff
@@ -92,8 +99,12 @@ def check_version_updated(files: List[Path], commit_range: str) -> bool:
...
     rcode = True
     for nasl_file in files:
-        if nasl_file.suffix != ".nasl" or not nasl_file.exists():
-            continue
+        if (
+            nasl_file.suffix != ".nasl"
+            or not nasl_file.exists()
+            or is_ignore_file(nasl_file, _IGNORE_FILES)
+        ):
+            return
```
cfi-gb, line 107:
```suggestion
            continue
```

cfi-gb: Fix in #557

PR #556, opened 2023-04-26
Add: troubadix-allowed-rev-diff tool

mbrinkhoff:
## What

Adds a tool to troubadix that checks the diff between separate revs conforms to given patterns.

## Why

Required for the release branch
Part of: VTD-1666

## Checklist

- [ ] Tests

### DIFF
--- a/troubadix/standalone_plugins/allowed_rev_diff.py
+++ b/troubadix/standalone_plugins/allowed_rev_diff.py
@@ -43,7 +43,7 @@ def parse_arguments() -> Namespace:
         type=str,
         default=DEFAULT_IGNORED_LINESTARTS,
         help="A list of line starts which will make the line be ignored. "
-        f"Default: {', '.join(DEFAULT_IGNORED_LINESTARTS)}",
+        "Default: %(default)s",
     )
 
     ignored_linestart_group.add_argument(
@@ -53,7 +53,7 @@ def parse_arguments() -> Namespace:
         default=DEFAULT_IGNORED_LINESTARTS,
         help="The file containing a list of line starts, "
         "separated by newlines, which will make the line be ignored. "
-        f"Default values used: {', '.join(DEFAULT_IGNORED_LINESTARTS)}",
+        "Default values used: %(default)s",
     )
 
     pattern_group = argument_parser.add_mutually_exclusive_group(required=True)
--- a/troubadix/standalone_plugins/allowed_rev_diff.py
+++ b/troubadix/standalone_plugins/allowed_rev_diff.py
@@ -144,7 +144,13 @@ def main() -> int:
 
     result = check_diff(diff.splitlines(), ignored_linestarts, patterns)
 
-    return 1 if result else 0
+    if result:
+        print("The following lines don't match the any pattern:")
+        print("\n".join(result))
+
+        return 1
+
+    return 0
 
 
 if __name__ == "__main__":
--- a/troubadix/standalone_plugins/allowed_rev_diff.py
+++ b/troubadix/standalone_plugins/allowed_rev_diff.py
@@ -38,15 +38,19 @@ def parse_arguments() -> Namespace:
     ignored_linestart_group.add_argument(
         "-i",
         "--ignored-linestarts",
+        dest="ignored_linestarts",
         nargs="*",
         type=str,
+        default=DEFAULT_IGNORED_LINESTARTS,
         help="A list of line starts which will make the line be ignored. "
         f"Default: {', '.join(DEFAULT_IGNORED_LINESTARTS)}",
     )
 
     ignored_linestart_group.add_argument(
         "--ignored-linestart-file",
-        type=Path,
+        type=read_ignored_linestarts,
+        dest="ignored_linestarts",
+        default=DEFAULT_IGNORED_LINESTARTS,
         help="The file containing a list of line starts, "
         "separated by newlines, which will make the line be ignored. "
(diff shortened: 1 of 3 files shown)


troubadix/standalone_plugins/allowed_rev_diff.py
```diff
@@ -0,0 +1,161 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
cfi-gb, line 16:
```suggestion
# SPDX-FileCopyrightText: 2023 Greenbone AG
#
# SPDX-License-Identifier: GPL-3.0-or-later
```

For new plugins maybe directly use the new header, see e.g. greenbone/pontos#681

troubadix/standalone_plugins/allowed_rev_diff.py
```diff
@@ -0,0 +1,161 @@
...
+) -> bool:
+    return any(
+        linestart
+        for linestart in ignored_linestarts
+        if line.startswith(linestart)
+    )
+
+
+def check_diff_line_matches_pattern(line: str, patterns: List[Pattern]) -> bool:
+    return any(pattern for pattern in patterns if pattern.match(line))
```
n-thumann, line 107:
```suggestion
    return any(pattern.match(line) for pattern in patterns)
```
We could simplify this a little but, because we don't care about the list comprehensions result anyways

troubadix/standalone_plugins/allowed_rev_diff.py
```diff
@@ -0,0 +1,161 @@
...
+
+    target = repo.commit(arguments.target)
+
+    merge_base = repo.merge_base(repo.commit(arguments.source), target)[0]
+
+    diff = repo.git.diff(target, merge_base, unified=0)
+
+    result = check_diff(diff.splitlines(), ignored_linestarts, patterns)
+
+    return 1 if result else 0
```
n-thumann, line 157: Can we have some basic output, if this isn't empty? I think this would ease debugging or configuring a correct RegEx in the first place:

```suggestion
    if result:
        print("The following lines don't match the any pattern:")
        print("\n".join(result))
        return 1

    return 0
```

troubadix/standalone_plugins/allowed_rev_diff.py
```diff
@@ -0,0 +1,161 @@
...
+    if not path:
+        return None
+
+    with open(path, "r", encoding="UTF-8") as file:
+        return [line[:-1] for line in file.readlines()]
+
+
+def read_patterns(path: Path) -> List[Pattern]:
+    with open(path, "r", encoding="UTF-8") as file:
+        return [re.compile(pattern[:-1]) for pattern in file.readlines()]
```
n-thumann, line 133: Is the `pattern[:-1]` to strip the line break? If so, using `.removesuffix("\n")` might be easier to understand

n-thumann (changes requested):

troubadix/standalone_plugins/allowed_rev_diff.py
mbrinkhoff: Done

troubadix/standalone_plugins/allowed_rev_diff.py
mbrinkhoff: True, fixed

troubadix/standalone_plugins/allowed_rev_diff.py
```diff
@@ -0,0 +1,161 @@
...
+
+    ignored_linestart_group = argument_parser.add_mutually_exclusive_group()
+
+    ignored_linestart_group.add_argument(
+        "-i",
+        "--ignored-linestarts",
+        nargs="*",
+        type=str,
+        help="A list of line starts which will make the line be ignored. "
+        "Default: ['diff ', 'index ', '--- ', '+++ ', '@@ ']",
```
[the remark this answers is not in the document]
n-thumann: I think it should work, if we let `argparse` take care of the default arguments stuff:
With `type=read_ignored_linestarts` and `dest="ignored_linestarts"` on `--ignored-linestart-file`, `arguments.ignored_linestarts` will always contain the expected data (either from CLI or from file or default values):

```
➜  troubadix git:(release_branch_diff_check_tool) ✗ poetry run troubadix-allowed-rev-diff -s x -t y -p z --ignored-linestart-file ignored_linestarts.txt
['foo', 'bar']
➜  troubadix git:(release_branch_diff_check_tool) ✗ poetry run troubadix-allowed-rev-diff -s x -t y -p z --ignored-linestarts foo bar                   
['foo', 'bar']
➜  troubadix git:(release_branch_diff_check_tool) ✗ poetry run troubadix-allowed-rev-diff -s x -t y -p z
['diff ', 'index ', '--- ', '+++ ', '@@ ']
```

Then we can also specify a default value for this argument (and probably just omit it for the file argument)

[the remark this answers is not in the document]
mbrinkhoff: I will give that a try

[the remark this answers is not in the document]
mbrinkhoff: works

PR #591, opened 2023-07-13
Change: Add 'package_unreliable' to the allowed qod_type values.

cfi-gb:
## What
The new `package_unreliable` qod_type had been introduced via
- AP-1928
- SC-506
- greenbone/gsa#3353
- greenbone/gvmd#1782
- greenbone/ospd-openvas#605
in our software stack but wasn't added to Troubadix in the past.

Note: #590 should be merged first before merging this so that both are included in the new release

## Why
Allow the already implemented new qod_type in VTs.

## References
N/A

### DIFF
--- a/troubadix/plugins/qod.py
+++ b/troubadix/plugins/qod.py
@@ -48,6 +48,7 @@ VALID_QOD_TYPES = [
     "remote_banner_unreliable",
     "executable_version_unreliable",
     "general_note",
+    "package_unreliable",
 ]
 
 


mbrinkhoff:
After discussing with Nicolas, QoD of `package_unreliable` will only be available for >=22.04, we should check for that (at least kinda) I assume ?

cfi-gb:
From https://forum.greenbone.net/t/greenbone-os-21-04-retired-final-patch-level-gos-21-04-26/14365:

> All GOS software packages and 3rd-party software below version 22.04 are **no longer supported** and **their compatibility** with the Greenbone Enterprise Feed or the Greenbone Community Feed **is no longer guaranteed**. Their functionality may cease or be limited at any time.

I don't think that we need any special handling for versions < 22.04 currently.

n-thumann (commented):
~Do we need to add a check if the NASL script is for 22.04 or later? Because IIRC `package_unreliable` isn't valid in 21.04 or earlier~ nvm, didn't saw Max comment

mbrinkhoff (approved):

PR #595, opened 2023-07-28
Improve encoding detection

n-thumann:
## What
This PR improves encoding detection by using charset-normalizer instead of chardet.

## Why
For example, chardet detects `geprüft` as ISO-8859-9:
```
>>> b'\x67\x65\x70\x72\xfc\x66\x74'.decode("latin_1")
'geprüft'
>>> chardet.detect(b'\x67\x65\x70\x72\xfc\x66\x74')
{'encoding': 'ISO-8859-9', 'confidence': 0.7300222957397617, 'language': 'Turkish'}
```
While this is technically correct, `geprüft` is also valid ISO-8859-1.

My first approach was to manually detect the encodings by calling e.g. `.decode("latin_1")` on the NASL content to check, if the content can be decoded with ASCII or Latin-1. If it can't (like UTF-8), it's an invalid encoding. This however turned out to be not as reliable as expected.

I therefore switched to charset-normalizer and pass our only two allowed encodings.
```
>>> charset_normalizer.from_bytes(b'\x67\x65\x70\x72\xfc\x66\x74', cp_isolation=["ascii", "latin_1"]).best().encoding
'latin_1'
```
Limiting the encodings is technically not necessary, but improves runtime (it only checks two instead of all encodings) and eases the usage for us (we don't need to search the result for our encodings). The tradeoff is that we can no longer print the guessed encoding. I also increase the threshold from 0.2 to 0.4 to keep trying harder on bigger, more chaotic looking files (this currently only affects `common/2016/gb_cisco_ios_cisco-sa-20100922-h323.nasl`). 

## References
Would have fixed https://github.com/greenbone/vulnerability-tests/pull/6272

## Test
Example from https://github.com/greenbone/vulnerability-tests/pull/6272:
```
➜  troubadix git:(main) poetry run troubadix -v --include-tests CheckEncoding --file ../vulnerability-tests/nasl/common/GSHB/GSHB_Kompendium.nasl
ℹ Start linting 1 files ... 
ℹ Checking common/GSHB/GSHB_Kompendium.nasl (1/1)
ℹ     Results for plugin check_encoding
×         VT uses a wrong encoding. Detected encoding is ISO-8859-7.
ℹ Time elapsed: 0:00:00.060085
× check_encoding                                          1        0
ℹ sum                                                     1        0
➜  troubadix git:(improve_encoding_detection) poetry run troubadix -v --include-tests CheckEncoding --file ../vulnerability-tests/nasl/common/GSHB/GSHB_Kompendium.nasl
ℹ Start linting 1 files ... 
ℹ Time elapsed: 0:00:00.035964
ℹ sum                                                     0        0
```

I also ran this Troubadix plugin against the whole `nasl/` directory with no issues reported.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -174,6 +174,90 @@ files = [
 ]
 
 [[package]]
+name = "charset-normalizer"
+version = "3.2.0"
+description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
+optional = false
+python-versions = ">=3.7.0"
+files = [
+    {file = "charset-normalizer-3.2.0.tar.gz", hash = "sha256:3bb3d25a8e6c0aedd251753a79ae98a093c7e7b471faa3aa9a93a81431987ace"},
+    {file = "charset_normalizer-3.2.0-cp310-cp310-macosx_10_9_universal2.whl", hash = "sha256:0b87549028f680ca955556e3bd57013ab47474c3124dc069faa0b6545b6c9710"},
+    {file = "charset_normalizer-3.2.0-cp310-cp310-macosx_10_9_x86_64.whl", hash = "sha256:7c70087bfee18a42b4040bb9ec1ca15a08242cf5867c58726530bdf3945672ed"},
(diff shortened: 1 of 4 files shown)


cfi-gb:
Please make sure to merge #596 before so that we have included both in the next release.

mbrinkhoff (approved):

PR #594, opened 2023-07-27
Require Python 3.9+ and update dependencies

n-thumann:
## What
This PR bumps the minimum required Python version to 3.9 or higher and updates this projects dependencies.

## Why
- The bump to Python 3.9+ is required because
  - The latest Greenbone autohooks require Python 3.8+
  - The latest `pontos` requires Python 3.9+
  - Not having the latest version of them also causes their dependencies to be outdated and not updatable
- Dependabot is still failing for multiple dependencies (see https://github.com/dependabot/dependabot-core/issues/7389): 
```
updater | Dependabot encountered '21' error(s) during execution, please check the logs for more details.
updater | +------------------------------------+
updater | |   Dependencies failed to update    |
updater | +--------------------+---------------+
updater | | isort              | unknown_error |
updater | | rich               | unknown_error |
updater | | zipp               | unknown_error |
updater | | exceptiongroup     | unknown_error |
updater | | markdown-it-py     | unknown_error |
updater | | pylint             | unknown_error |
updater | | black              | unknown_error |
updater | | setuptools         | unknown_error |
updater | | platformdirs       | unknown_error |
updater | | httpcore           | unknown_error |
updater | | typed-ast          | unknown_error |
updater | | httpx              | unknown_error |
updater | | click              | unknown_error |
updater | | certifi            | unknown_error |
updater | | rfc3986            | unknown_error |
updater | | semver             | unknown_error |
updater | | astroid            | unknown_error |
updater | | typing-extensions  | unknown_error |
updater | | dill               | unknown_error |
updater | | anyio              | unknown_error |
updater | | importlib-metadata | unknown_error |
updater | +--------------------+---------------+
```
- It resolves https://github.com/greenbone/troubadix/security/dependabot/3
- It fixes a few pylint warnings that popped up after this update:
```
pylint: Command line or configuration file:1: UserWarning: Specifying exception names in the overgeneral-exceptions option without module name is deprecated and support for it will be removed in pylint 3.0. Use fully qualified name (maybe 'builtins.Exception' ?) instead.
************* Module /home/runner/work/troubadix/troubadix/.pylintrc
.pylintrc:1:0: E0015: Unrecognized option found: no-space-check (unrecognized-option)
************* Module troubadix.plugins.spelling
troubadix/plugins/spelling.py:75:0: W1404: Implicit string concatenation found in call (implicit-str-concat)
************* Module tests.plugins.test_cve_format
[...]

PR #609, opened 2023-09-06
Change: Downgrade validators to 0.20.0

n-thumann:
## What
This PR downgrades `validators` to 0.20.0 after a version bump to a buggy version via #594.

## Why
To get rid off false positives. I checked every version, starting from the latest (0.22.0) and 0.20.0 was the first one with no false positives.

## References
This can be revered as soon as https://github.com/python-validators/validators/issues/296 is solved.

## Checklist
I ran the Troubadix against the affected NASL script and confirmed that there no longer is a false positive reported.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -310,6 +310,17 @@ files = [
 colorama = {version = "*", markers = "platform_system == \"Windows\""}
 
 [[package]]
+name = "decorator"
+version = "5.1.1"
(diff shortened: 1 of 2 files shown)


cfi-gb (approved):
While i don't have much knowledge on this Python dependency topic the change looks / sane when comparing to the previous 0.20.0 state from #594

PR #620, opened 2023-09-14
Change: Various updates for the conventional commits handling

cfi-gb:
## What
- Update conventional-commits.yml to work with bot PRs
- Use default config for conventional commits

Use the same code / content as in e.g. https://github.com/greenbone/pontos/blob/main/.github/workflows/conventional-commits.yml which includes the updates from:
- greenbone/pontos#822
- greenbone/pontos#785

and also do the same as greenbone/pontos#793 to use the default config for conventional commits.

## Why

#616 is now prefixing all Dependabot PRs with `Deps:` to have them included in the changelog. But it seems the conventional-commits workflow wasn't able do identify these (probably because it wasn't able to get the data / info from the PR) and  thus only reported a `No conventional commits found.` like seen here:

https://github.com/greenbone/troubadix/pull/618#issuecomment-1718023674

while there actually was a conventional commit included in that PR.

This PR should make this to work.

## References

None

### DIFF
--- a/.github/workflows/conventional-commits.yml
+++ b/.github/workflows/conventional-commits.yml
@@ -1,15 +1,16 @@
 name: Conventional Commits
 
 on:
-  pull_request:
+  pull_request_target:
+
+permissions:
+  pull-requests: write
+  contents: read
 
 jobs:
   conventional-commits:
     name: Conventional Commits
     runs-on: ubuntu-latest
-    permissions:
-      pull-requests: write
-      contents: read
     steps:
       - name: Report Conventional Commits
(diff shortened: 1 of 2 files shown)


bjoernricks:
To make that work you need to use the default config for evaluating the changelog prefix by deleting the changelog.toml file from the repo

cfi-gb:
> To make that work you need to use the default config for evaluating the changelog prefix by deleting the changelog.toml file from the repo

Good hint, thanks a lot. I have updated this PR accordingly :+1:

mbrinkhoff (approved):

PR #678, opened 2024-03-11
Add: Extend date verification for last modification tag

NiklasHargarterGb:
## What

Use the datetime module to parse and validate the retrieved last modification date. Added checks for general structure, containing only single date, weekday matching the date.

## Why

The last modification date is only checked for correct length, not for syntax and values.

## References

[VTOPS-128](https://jira.greenbone.net/browse/VTOPS-128)

## Checklist

- [x] Tests

### DIFF
--- a/troubadix/plugins/script_version_and_last_modification_tags.py
+++ b/troubadix/plugins/script_version_and_last_modification_tags.py
@@ -129,6 +129,41 @@ class CheckScriptVersionAndLastModificationTags(FileContentPlugin):
                 file=nasl_file,
                 plugin=self.name,
             )
+        else:
+            date_str = match_last_modified.group("value")
+            format_left = "%Y-%m-%d %H:%M:%S %z "
+            format_right = "(%a, %d %b %Y)"
+            try:
+                date_left = datetime.datetime.strptime(
+                    date_str[:26], format_left
+                )
+                date_right = datetime.datetime.strptime(
+                    date_str[26:], format_right
+                )
+                week_day_parsed = date_right.strftime("%a")
+            except ValueError:
+                yield LinterError(
+                    "False or incorrectly formatted modification_date.",
+                    file=nasl_file,
+                    plugin=self.name,
+                )
+                return
+            week_day_str = date_str[27:30]
+            # Wed, 29 Nov 2017
+            if date_left.date() != date_right.date():
+                yield LinterError(
+                    "The modification_date consists of two different dates.",
+                    file=nasl_file,
+                    plugin=self.name,
+                )
+            # Check correct weekday
+            elif week_day_str != week_day_parsed:
+                yield LinterError(
+                    f"Wrong day of week. Please change it from '{week_day_str}"
+                    f"' to '{week_day_parsed}'.",
+                    file=nasl_file,
+                    plugin=self.name,
+                )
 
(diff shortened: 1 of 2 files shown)


troubadix/plugins/script_version_and_last_modification_tags.py
```diff
@@ -129,6 +129,41 @@ def check_content(
...
+            # Wed, 29 Nov 2017
+            if date_left.date() != date_right.date():
+                yield LinterError(
+                    "The modification_date consists of two different dates.",
+                    file=nasl_file,
+                    plugin=self.name,
+                )
+            # Check correct weekday
+            elif week_day_str != week_day_parsed:
+                yield LinterError(
```
n-thumann, line 161: Maybe `return` here too instead of `elif`? But that's just a personal preference :)

troubadix/plugins/script_version_and_last_modification_tags.py
```diff
@@ -129,6 +129,41 @@ def check_content(
...
+            except ValueError:
+                yield LinterError(
+                    "False or incorrectly formatted modification_date.",
+                    file=nasl_file,
+                    plugin=self.name,
+                )
+                return
+            week_day_str = date_str[27:30]
+            # Wed, 29 Nov 2017
+            if date_left.date() != date_right.date():
```
n-thumann, line 153: Why can't we compare the two `datetime` objects directly?

n-thumann (changes requested):

troubadix/plugins/script_version_and_last_modification_tags.py
NiklasHargarterGb: Because the date object contains the same date twice, and the format then contains capture targets like the year %y twice. And this library only works with parsing a single target from a date object once using capture groups. It then complains that capture groups are being redefined. [related pull request explaining why it doesn't work](https://github.com/python/cpython/issues/48680)

troubadix/plugins/script_version_and_last_modification_tags.py
```diff
@@ -129,6 +129,41 @@ def check_content(
                 file=nasl_file,
                 plugin=self.name,
             )
+        else:
+            date_str = match_last_modified.group("value")
```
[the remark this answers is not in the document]
n-thumann: Done in 9d6c32dba64917df1e278f176bcf5ed0d41e8af6

troubadix/plugins/script_version_and_last_modification_tags.py
n-thumann: I actually meant why we can't just `date_left == date_right`, but missed that this obviously cannot work, because the left contains hours, minutes, ..., while the right doesn't. Nevermind then :)

troubadix/plugins/script_version_and_last_modification_tags.py
NiklasHargarterGb: Done in 9d6c32dba64917df1e278f176bcf5ed0d41e8af6

n-thumann (approved):
I ran this plugin again all VTs in `nasl/common`. There were three VTs reported with an incorrect date with all of them being actually faulty. Those three were previously not found.
The runtime increase from ~10 sec. to ~12 seconds.

PR #679, opened 2024-03-13
Add: Add version syntax verification

NiklasHargarterGb:
## What

Use the datetime module to validate the retrieved value of the version tag.

## Why

The version value is only checked for correct length, not for syntax.

## References

[VTOPS-128](https://jira.greenbone.net/browse/VTOPS-128)

## Checklist

- [x] Tests


troubadix/plugins/script_version_and_last_modification_tags.py
```diff
@@ -94,6 +94,17 @@ def check_content(
                 file=nasl_file,
                 plugin=self.name,
             )
+        else:
+            version_str = version_match.group("value")
+            format_version = "%Y-%m-%dT%H:%M:%S%z"
```
n-thumann, line 99:
```suggestion
                plugin=self.name,
            )
            return

        version_str = version_match.group("value")
        format_version = "%Y-%m-%dT%H:%M:%S%z"
```
Maybe we can return early to avoid some indentation / one level of nesting (`yield` in `except` in `else`)

NiklasHargarterGb: This would skip the last_modification checks it currently performs, since it has already determined that the version is missing. The existing tests required it to check both in the same run, even if version failed.

tests/plugins/test_script_version_and_last_modification_tags.py
```diff
@@ -137,6 +137,29 @@ def test_missing_script_version(self):
...
+        nasl_file = Path(__file__).parent / "test.nasl"
+        content = (
+            '  script_tag(name:"cvss_base", value:"4.0");\n'
+            '  script_tag(name:"summary", value:"Foo Bar.");\n'
+            '  script_tag(name:"solution_type", value:"VendorFix");\n'
+            '  script_tag(name:"solution", value:"meh");\n'
+            '  script_version("TTTTT07-19T12:32:02+0000");\n'
+            '  script_tag(name: "last_modification", value: "2021-07-19 '
+            '12:32:02 +0000 (Mon, 19 Jul 2021)");\n'
+        )
```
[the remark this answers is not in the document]
NiklasHargarterGb: implemented in 8e15359327baab162f5ce09245099af5065de67a

troubadix/plugins/script_version_and_last_modification_tags.py
n-thumann: Ah, true.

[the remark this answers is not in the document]
NiklasHargarterGb: implemented in 3fc85f1a914a7f2983d5b0713827e87cb4957b4e

PR #683, opened 2024-03-21
Fix: Codecov in workflow

n-thumann:
## What
This PR fixes Codecov in our workflow.

## Why
To have Codecov working again.
For whatever reason we already have `secrets.CODECOV_TOKEN` present, but we're not using it.

## References
This kind of error has been first introduced in https://github.com/greenbone/troubadix/pull/668, but can't really tell why. Also, the error got shadowed by another one, because in the CI runs prior to that one it also failed due to a different error.

## Checklist

- [ ] Tests

### DIFF
--- a/.github/workflows/ci-python.yml
+++ b/.github/workflows/ci-python.yml
@@ -67,5 +67,5 @@ jobs:
       - name: Upload coverage to codecov.io
         uses: codecov/codecov-action@v4
         with:
-          token: ${{ inputs.token }}
+          token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: true


cfi-gb:
Only a remark: Do we need a conventional commit for this change? Asking because with this it will end up as a changelog entry for Troubadix on the next release here:

https://github.com/greenbone/troubadix/releases

while this is completely not about Troubadix but just some internal repo maintenance.

n-thumann:
> Only a remark: Do we need a conventional commit for this change? Asking because with this it will end up as a changelog entry for Troubadix on the next release here:

AFAIK from other projects it's not uncommon to have also changes that are not directly related the actual code in the changelog 🤔 I'd don't mind keeping it there, but I leave that up to discussion

cfi-gb:
IMHO any changes not directly related to the software shipped with a repository shouldn't end up in the changelog of that repository. This is something i personally really dislike in these other projects:

As a user of the software i have no interest in anything other then actual changes to the software, a changelog should purely contain only changes to the software itself and shouldn't include anything development related.

PR #706, opened 2024-05-14
Add: Use external toml file for plugin configuration

ghost:
## What
Use an external toml file to store plugin configurations.
Rewrite http_links_in_tags plugin to use external config for exclusions as a demonstration of how other plugins can use it.

The location of the file is specified by the --plugins-config-file command-line argument added to the parser.
This argument has a default value of 'troubadix_config.toml'. This is the same as the config file in the vt repo.
This makes this release compatible with all old workflows and aliases that use Troubadix, as long as the new config file is present.
Otherwise, a different file must be specified, as Troubadix will no longer be able to run without a valid config file. 
If no plugin is being run that uses external configuration, an empty but existing file is sufficient to be valid.
This is done to make it less likely that Troubadix will run without a configuration when it should have one.

This file must have an entry with the same name as the plugins Plugin.name attribute for each plugin that is executed during the Troubadix run that uses the new system.

```toml
[Plugin.name]
exclusions = ["foo", "bar"]
```
Plugins that are not executed or that do not use the new system do not need an entry in the toml file.

The values of the plugin entries can be any valid Toml structure that can be parsed into Python types.
This should allow the configuration of most plugins.
For example, http_tags_in_link just wants a list of exclusions. Which maps 1 to 1 from toml to python.

Plugins must use a class attribute to indicate if they depend on an external configuration. If nothing is set, as in existing plugins, the default of False is used. And then none of the new changes will apply to this plugin. Its function remains the same.

Plugins that use external config are passed the entire config, and are responsible for validating and extracting only the parts they want. They can then access the config via an instance attribute `config

## Why
The purpose is to unify the configuration for all Troubadix plugins that are use case dependent / come from outside Troubadix.

Exclusion changes that are necessary for VT development, but do not change the Troubadix plugins, can then be part of the same PR that introduces the VT, and not require an additional Troubadix PR anymore.

### Rollout Step1 
A toml file must be placed in the VT repository that contains all configuration for plugins that use the new file.
### Rollout Step2
If a config file from a location other than the default location is to be used for QA, the troubadix invocation must be changed to reflect this.
### Changes to config during normal VT development
Changes to the configuration of plugins that use the new system must be made in the vt repo toml file, not in Troubadix. 
Changes to the config of plugins that have not yet been converted still need a Troubadix PR.

## Converting existing plugins / Creating new Plugin with external config
Set class attribute `requires_external_config = True`
Override `validate_plugin_config` of the 'Plugin' base class.
Access config via `self.config["key"]`.
## References
VTOPS-115

## Checklist

- [ ] Tests


troubadix/plugins/http_links_in_tags.py
```diff
@@ -116,62 +118,8 @@ def contains_nvd_mitre_link_in_xref(self) -> Iterator[LinterResult]:
...
-            "https:// ",
-            "such as 'http://:80'",
-            "<http://localhost/moodle/admin/>",
-            "https://username:password@proxy:8080",
-            "sun.net.www.http.KeepAliveCache",
-            "www.foo.com",
-        ]
-
+    def check_to_continue(self, http_link_match_group: str) -> bool:
+        exclusions = self.context.plugin_config.get("exclusions", [])
```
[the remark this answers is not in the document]
NiklasHargarterGb: You mean that the plugin will not work completely if it does not get an exclusion list from the config file?
Then the config_file is no longer optional, but fine by me.
Or only do the exclusion check if it has a proper exclusion list?

troubadix/runner.py
```diff
@@ -91,6 +93,8 @@ def _check_file(self, file_path: Path) -> FileResults:
         )
 
         for plugin_class in self.plugins.file_plugins:
+            plugin_config = self.plugins_config.get(plugin_class.name, {})
+            context.plugin_config = plugin_config
```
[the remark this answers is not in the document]
NiklasHargarterGb: The intention was to have only the specific plugin config for the plugin running in the file context, since the same context is used by multiple plugins. And then it could only be added at the point where the plugin is decided, but the context has already been created. When editing the context  __init__ I would then pass the entire config dict and then in the plugin itself it would have to get the specific config from all configs.

troubadix/argparser.py
```diff
@@ -28,14 +28,14 @@
...
-    if directory_path.exists() and not directory_path.is_dir():
+    if not directory_path.is_dir():
         raise ValueError(f"{string} is not a directory.")
     return directory_path
 
 
 def file_type(string: str) -> Path:
     file_path = Path(string)
-    if file_path.exists() and not file_path.is_file():
+    if not file_path.is_file():
```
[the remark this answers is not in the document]
NiklasHargarterGb: Added comments that passing non existent paths is intended and should not be changed.

troubadix/troubadix.py
```diff
@@ -163,6 +164,12 @@ def main(args=None):
         first_file = files[0].resolve()
         root = get_root(first_file)
 
+    if parsed_args.plugins_config_file:
+        with open(parsed_args.plugins_config_file, "rb") as file:
+            plugins_config = tomli.load(file)
+    else:
+        plugins_config = {}
```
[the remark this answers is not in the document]
NiklasHargarterGb: That was done so that the file is seen as optional and Troubadix can be run without it.
I will remove the default values that keep it from crashing. I will set it as a required parameter in argparse.

troubadix/runner.py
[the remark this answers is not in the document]
NiklasHargarterGb: Plugin classes `__init__` can be changed to take the plugin_config as parameter directly and be then seperate from the file_context.

troubadix/argparser.py
```diff
@@ -28,13 +28,17 @@
...
+    # DO NOT CHANGE!
     if directory_path.exists() and not directory_path.is_dir():
         raise ValueError(f"{string} is not a directory.")
     return directory_path
 
 
 def file_type(string: str) -> Path:
     file_path = Path(string)
+    # intentionally passes filepaths that don't exist as valid
+    # DO NOT CHANGE!
```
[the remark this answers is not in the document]
NiklasHargarterGb: file_type is used for actual files and for paths that should be files at some point, such as log files created when they do not yet exist.

For directory_type, I cannot tell you the reason. I cannot find the same case as above, where a non-existent directory is created on demand.

Not really sure how to test for this. Grepped for all its uses. directory_type is only used to pass a list of dirs to parse and the root directory. For dirs, it is practically ignored if they do not exist but are not created. Passing a non-existent root can cause errors, as sometimes files will be checked if they are relative to it, but again no directory will be created. So maybe this could be changed to not pass non-existent directories.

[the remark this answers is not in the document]
NiklasHargarterGb: done in 69c69b2e07ce9db14b535d05736360afe80b0ff9

troubadix_config.toml
```diff
@@ -0,0 +1,55 @@
+title = "Troubadix Ignore File"
+
+[check_http_links_in_tags]
+description = "Strings that should be ignored because they contain a valid URL type in a tag"
+exclusions = [
+            "The payloads try to open a connection to www.google.com",
```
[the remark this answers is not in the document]
NiklasHargarterGb: Yes, indentation is wrong. I now use Taplo for formatting. (which is also used by the VS Code Extension 'Even Better TOML)
69c69b2e07ce9db14b535d05736360afe80b0ff9

troubadix/plugins/http_links_in_tags.py
```diff
@@ -17,15 +17,41 @@
...
+    LinterResult,
+)
 
 
 class CheckHttpLinksInTags(FilePlugin):
     name = "check_http_links_in_tags"
+    mandatory_config_keys = ["exclusions"]
+    require_external_config = True
+
+    def validate_and_extract_plugin_config(self, config: dict) -> dict:
```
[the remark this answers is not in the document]
NiklasHargarterGb: i went with extract calling validate. Extract is still called by the init method of the base Plugin class. Plugins now only need to overwrite the validate method.
Done in 69c69b2e07ce9db14b535d05736360afe80b0ff9

pyproject.toml
```diff
@@ -35,6 +35,7 @@ chardet = ">=4,<6"
 validators = "0.20.0"
 gitpython = "^3.1.31"
 charset-normalizer = "^3.2.0"
+tomli = "^2.0.1"
```
n-thumann, line 38:
```suggestion
tomli = { version = "^2.0.1", python = "<3.11" }
```
Python 3.11+ comes with a TOML parser included, so we could do this and then something like
```
if (version_info.major, version_info.minor) >= (3, 11):
   import tomllib
    tomllib.load(...)
else:
   import tomli
   tomli.load(...)
```

NiklasHargarter: implemented in ec62dae9430946e354cdc76391d47e0917d8dea2.
But I use a slightly different structure because otherwise my lsp complains that the else part is unreachable.

troubadix_config.toml
n-thumann, line 1: Do we really want to add this to both the Troubadix and vulnerability-tests repo? This will likely lead to the same fate like the codespell files requiring to get sync'ed manually / getting out of sync :/
Or it this only temporary for the rollout to assure that stuff won't break?

troubadix_config.toml
```diff
@@ -0,0 +1,55 @@
+# Use Taplo / VS Code Even Better TOML for formatting
+
+title = "Troubadix Plugin Config File"
+
+[check_http_links_in_tags]
+description = "Strings that should be ignored because they contain a valid URL type in a tag"
```
n-thumann, line 6:
```suggestion
[check_http_links_in_tags]
```
Those two aren't used, right?

troubadix_config.toml
NiklasHargarter: Its only temporary / used as a demonstration of what the structure should be. I would probably keep a functional but empty file here so that troubadix can run without the vt repo, which is private.

troubadix_config.toml
NiklasHargarter: are not used in the code, functionally just comments. 
The purpose of the description fields was to make the otherwise not very understandable list of random values make more sense. Basically, give a description of what this particular plugin configuration wants and in what format. This way, if the file is only in the VT repo, developers can make adjustments to it without having to look in the plugin's source code.
This can also be done as comments rather than actual toml values. I just feel like the way comments are grayed out makes them less good as nice readable documentation.

n-thumann: I'd be in favor of using comments instead, because if a value isn't actually used anywhere, we shouldn't declare it in the first place and misuse them as comments 🤔

troubadix_config.toml
n-thumann: > I would probably keep a functional but empty file here so that troubadix can run without the vt repo, which is private.

An empty config just to keep something working sounds a little odd to me :D

If we want to keep it functional without the user providing a config I would either:
1. Make the config optional and skip the check with a warning (sounds dangerous, because a missing config can stay unnoticed)
2. Still fail hard, but add some guidance for the user how to resolve it, e.g. `Please provide a configuration or disable the plugin via --exclude-tests {{self.name}}` in `extract_plugin_config`.

troubadix/argparser.py
```diff
@@ -230,6 +235,17 @@ def parse_args(
         help="Don't print the statistic",
     )
 
+    parser.add_argument(
+        "--plugins-config-file",
+        type=file_type,
+        default="troubadix_config.toml",
```
n-thumann, line 241:
```suggestion
        default="troubadix.toml",
```
What do you think about this? It aligns a little more with other configs like `poetry.toml` or `pyproject.toml`, `Cargo.toml`

NiklasHargarter: sounds good

troubadix_config.toml
NiklasHargarter: done in de8d0b7e8b07473600cdfdf1ef6ad8cdf102b88e

troubadix_config.toml
NiklasHargarter: implemented a way to run with no config if no plugin needs one in https://github.com/greenbone/troubadix/pull/726

PR #710, opened 2024-06-03
Fix: Red Hat spelling

n-thumann:
## What
This PR fixes the spelling of `Red Hat`.

## Why
Because we use the correct spelling `Red Hat` instead of `RedHat` in the upcoming VTs. Troubadix will fail otherwise.

## Checklist

- [x] Tests


cfi-gb:
For clarification / just that it is clear that the family isn't getting renamed in the future:

It seems there was a `RedHat` typo since the beginning / introduction of this plugin and there was never a `RedHat` family.

PR #718, opened 2024-06-19
Add: standalone plugin for deprecating VTs

amy-gb:
## What

Add a standalone plugin to deprecate VTs. The script takes either a single file or a directory of files, and writes the updated file to a new location.

The changes the script should accomplish are (from the relevant internal Confluence page):
1) add a note to the summary of why the VT is deprecated
2) remove all code after the description block
3) add the early exit (66) to the end
4) add the deprecated=True tag
5) remove unneeded script tags

## Why

For the Red Hat deployment we need to deprecate hundreds of VTs. Therefore it makes sense to have a script that does this

## References

VTA-531

## Checklist

- [x] Tests

### DIFF
--- a/tests/standalone_plugins/test_deprecate_vts.py
+++ b/tests/standalone_plugins/test_deprecate_vts.py
@@ -1,0 +1,160 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+# SPDX-FileCopyrightText: 2024 Greenbone AG
+# pylint: disable=line-too-long
+import unittest
+from pathlib import Path
+from tests.plugins import TemporaryDirectory
+
+from troubadix.standalone_plugins.deprecate_vts import (
+    deprecate,
+    parse_args,
+    DeprecatedFile,
+    get_summary,
+    finalize_content,
+    update_summary,
+    get_files,
+)
+
+
+class ParseArgsTestCase(unittest.TestCase):
+    def test_parse_args(self):
+        testfile = "testfile.nasl"
+        output_path = "attic/"
+        input_path = "nasl/common"
+        oid_mapping_file = "oid_file.py"
+
+        args = parse_args(
+            [
+                "--file",
+                str(testfile),
+                "--output-path",
+                output_path,
+                "--oid-mapping-path",
+                oid_mapping_file,
+                "--input-path",
+                input_path,
+            ]
+        )
+        self.assertEqual(args.file, Path(testfile))
+        self.assertEqual(args.output_path, output_path)
+        self.assertEqual(args.oid_mapping_path, oid_mapping_file)
+        self.assertEqual(args.input_path, input_path)
+
+
+NASL_CONTENT = (
+    '...if(description)\n{\n  script_oid("1.3.6.1.4.1.25623.1.0.910673");'
+    '\n  script_version("2024-03-12T14:15:13+0000");'
+    '\n  script_name("RedHat: Security Advisory for gd (RHSA-2020:5443-01)");'
+    '\n  script_family("Red Hat Local Security Checks");\n  script_dependencies("gather-package-list.nasl");'
+    '\n  script_mandatory_keys("ssh/login/rhel", "ssh/login/rpms", re:"ssh/login/release=RHENT_7");'
+    '\n\n  script_xref(name:"RHSA", value:"2020:5443-01");\n  script_xref(name:"URL", value:"https://www.redhat.com/archives/rhsa-announce/2020-December/msg00044.html");'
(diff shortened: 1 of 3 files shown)


cfi-gb:
Note: I have redacted an internal link from the description.

tests/standalone_plugins/test_deprecate_vts.py
```diff
@@ -0,0 +1,164 @@
...
+    def test_update_summary_no_oid_match(self):
+        file = DeprecatedFile(
+            name="testfile.nasl",
+            full_path=Path("dir/testfile.nasl"),
+            content=NASL_CONTENT,
+        )
+        result = update_summary(file)
+        expected = (
+            "This VT has been deprecated.The remote host is missing an update for the 'gd'\n "
+            " package(s) announced via the RHSA-2020:5443-01 advisory."
```
cfi-gb, line 130: I would suggest to add the deprecation notice at the end of the summary tag and just use a more generic text like e.g.:

> The remote host is missing an update for the 'gd' package(s) announced via the RHSA-2020:5443-01 advisory.
>
> Note: This VT has been deprecated and replaced by a Notus scanner based one.

or similar.

troubadix/standalone_plugins/deprecate_vts.py
```diff
@@ -0,0 +1,264 @@
...
+
+@dataclass
+class DeprecatedFile:
+    name: str
+    full_path: Path
+    content: str
+
+
+# CHANGE AS NEEDED
+FILENAME_REGEX = re.compile(r"gb_rhsa_2021")
```
cfi-gb, line 33: Wouldn't it make sense to make this available as a command line argument?

troubadix/standalone_plugins/deprecate_vts.py
```diff
@@ -0,0 +1,264 @@
...
+    new_oid = reverse_mapping.get(old_oid)
+    return new_oid
+
+
+def finalize_content(content: str) -> str:
+    """Update the content field of the nasl script by adding the
+    deprecated tag and removing the extra content."""
+    content_to_keep = content.split("exit(0);")[0]
+    return content_to_keep + (
+        "script_tag(name: 'deprecated', value: TRUE);"
```
cfi-gb, line 90: The usual / most common coding style of these is:

```
script_tag(name: 'deprecated', value: TRUE);
```

1. Double quotes instead of single ones
2. No space after the `:`

troubadix/standalone_plugins/deprecate_vts.py
```diff
@@ -0,0 +1,264 @@
...
+        help="Path where the deprecated files should be written to.",
+    )
+    parser.add_argument(
+        "-f",
+        "--file",
+        metavar="<file>",
+        nargs="?",
+        default=None,
+        type=file_type,
+        help="single file to deprecate",
```
cfi-gb, line 220:
```suggestion
        help="Single file to deprecate",
```

tests/standalone_plugins/test_deprecate_vts.py
cfi-gb: Personally i would be absolutely fine to just use a general more standardized format for all Notus relevant deprecated LSCs, a user should be clever enough to add e.g. the Advisory-ID into the SecInfo search to find the replacements.

Note that such "Notus scanner based one" text should only be added if that applies to the deprecated VT.

troubadix/standalone_plugins/deprecate_vts.py
```diff
@@ -0,0 +1,265 @@
...
+
+@dataclass
+class DeprecatedFile:
+    name: str
+    full_path: Path
+    content: str
+
+
+# CHANGE AS NEEDED
+FILENAME_REGEX = re.compile(r"gb_")
```
cfi-gb, line 33: As mentioned in a now "Outdated" comment:

Wouldn't it make sense to use a command line argument for this case instead so that the file itself doesn't need to be touched?

tests/standalone_plugins/test_deprecate_vts.py
cfi-gb: Sounds great :+1:

troubadix/standalone_plugins/deprecate_vts.py
```diff
@@ -0,0 +1,249 @@
...
+    ScriptTag,
+    SpecialScriptTag,
+)
+
+DEPRECATIONS = {
+    "notus": "and replaced by a Notus scanner based one.",
+    "merged": "because it has been merged into a different VT.",
+    "defunct": "and is therefore no longer functional.",
+    "duplicate": "as a duplicate.",
+}
```
n-thumann, line 23: If you have an idea off the top of your head how to use an Enum instead, I'd prefer that. IIRC Enums and argparse are a bit nasty, so feel free to resolve this, if it can't be implemented quickly.

n-thumann (approved):

PR #751, opened 2024-09-09
Add: plugin for finding excess space before dot in some tags

NiklasHargarter:
## What
Adds a plugin to check script_tags that use full sentences in their value for excess white spaces before sentence terminating dots.
fix method is implemented to automatically remove the preceding excess white spaces. 

21.04 and 22.04 are ignored because they are generated.
Other entries in the ignore list use whitespace in various ways that have excess whitespace before dots, but on purpose, as used for beginning entries in an enumeration.

## References
Jira VTOPS-229

## Checklist

- [x] Tests

currently finds 121 occurrences in VT repo,

### DIFF
--- a/troubadix/plugins/spaces_before_dots.py
+++ b/troubadix/plugins/spaces_before_dots.py
@@ -2,7 +2,6 @@
 # SPDX-FileCopyrightText: 2024 Greenbone AG
 import re
 from collections.abc import Iterator
-from operator import itemgetter
 from pathlib import Path
 
 from troubadix.helper import CURRENT_ENCODING
@@ -33,8 +32,8 @@ TAGS = [
 PATTERN = re.compile(r"\s+\.(\s|$)")
 IGNORE = [
     # 21.04 and 22.04 are generated and should not be touched manually
-    "21.04",
-    "22.04",
+    "21.04/",
+    "22.04/",
     # uses dots for beginning of entry in enumeration
     "common/2008/debian/deb_246.nasl",
     "common/2008/debian/deb_266.nasl",
@@ -85,21 +84,23 @@ class CheckSpacesBeforeDots(FileContentPlugin):
         for tag in TAGS:
             pattern = get_script_tag_pattern(tag)
             match = pattern.search(file_content)
-            if match:
-                value = match.group("value")
-                value_start = match.start("value")
+            if not match:
+                continue
 
-                for excess_match in PATTERN.finditer(value):
-                    whitespace_pos = excess_match.start() + value_start
-                    self.matches.append((whitespace_pos, excess_match.group()))
-                    fullmatch = match.group()
-                    yield LinterWarning(
-                        f"value of script_tag {match.group('name')} has at least"
-                        " one occurence of excess whitespace before a dot:"
-                        f"\n '{fullmatch}'",
-                        file=nasl_file,
-                        plugin=self.name,
-                    )
+            value = match.group("value")
+            value_start = match.start("value")
+
+            for excess_match in PATTERN.finditer(value):
+                whitespace_pos = excess_match.start() + value_start
+                self.matches.append((whitespace_pos, excess_match.group()))
+                fullmatch = match.group()
+                yield LinterWarning(
+                    f"value of script_tag {match.group('name')} has at least"
+                    " one occurence of excess whitespace before a dot:"
+                    f"\n '{fullmatch}'",
+                    file=nasl_file,
+                    plugin=self.name,
+                )
 
     def fix(self) -> Iterator[LinterResult]:
 
@@ -107,7 +108,7 @@ class CheckSpacesBeforeDots(FileContentPlugin):
             return
 
         # Sort matches by position, descending order to avoid messing up indices during replacement
-        self.matches.sort(reverse=True, key=itemgetter(0))
+        self.matches.sort(reverse=True)
 
         file_content = self.context.file_content
         for pos, match_str in self.matches:


tests/plugins/test_spaces_before_dots.py
```diff
@@ -0,0 +1,56 @@
...
+from troubadix.plugins.spaces_before_dots import CheckSpacesBeforeDots
+
+
+class TestSpacesBeforeDots(PluginTestCase):
+
+    def test_ok(self):
+        nasl_file = Path("/some/fake/directory/test.nasl")
+        content = """
+              script_tag(name:"summary", value:"Foo Bar.");
+              script_tag(name:"solution", value:"meh.");
```
cfi-gb, line 16:
```suggestion
              script_tag(name:"insight", value:"foo bar ...");
              script_tag(name:"solution", value:"meh.");
```

Might be valid so i would not report these

tests/plugins/test_spaces_before_dots.py
```diff
@@ -0,0 +1,56 @@
...
+        plugin = CheckSpacesBeforeDots(fake_context)
+        results = list(plugin.run())
+        self.assertEqual(len(results), 6)
+        self.assertEqual(
+            results[0].message,
+            "value of script_tag summary has a excess space before the dot:\n"
+            " 'script_tag(name:"
+            '"summary", value:"Foo Bar .");'
+            "'",
+        )
```
cfi-gb, line 47: As shown in VTOPS-229 i have also seen something like this "live" / happening which should also be reported:

```
  script_tag(name:"affected", value:"Adobe Reader 2017 before 2017.011.30166
 .");
```

troubadix/plugins/spaces_before_dots.py
```diff
@@ -0,0 +1,47 @@
...
+        This plugin checks for excess space before the dot
+        in script_tags that have full sentence values
+        """
+        if nasl_file.suffix == ".inc":
+            return
+        for tag in TAGS:
+            pattern = get_script_tag_pattern(tag)
+            match = pattern.search(file_content)
+            if match:
+                s = match.group("value")
```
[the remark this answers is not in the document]
NiklasHargarter: there are many valid occurences of spaces before dots liike with .net and other words that start with a '.'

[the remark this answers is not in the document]
NiklasHargarter: if going with regex and finding:
1. dots enclosed in whitespace "foo . bar"
2. whitespace before end of string: "foo ." "foo\n"
I could probably use '\s\.\s|\s\.$'

tests/plugins/test_spaces_before_dots.py
NiklasHargarter: tripple dots will not trigger a report. But included as a test in 9a8f9bdcf577ce0a5fc421742bfc464adbcfc1e4 to make sure.

tests/plugins/test_spaces_before_dots.py
NiklasHargarter: Completely overlooked that newline should be included. Now it covers all white space characters and not only spaces. Done in 9a8f9bdcf577ce0a5fc421742bfc464adbcfc1e4

troubadix/plugins/spaces_before_dots.py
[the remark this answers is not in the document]
NiklasHargarter: Changed to the regex above in 9a8f9bdcf577ce0a5fc421742bfc464adbcfc1e4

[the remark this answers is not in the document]
NiklasHargarter: done in 9a8f9bdcf577ce0a5fc421742bfc464adbcfc1e4

cfi-gb (commented):
The existing test cases LGTM, thanks a lot for taking care :+1: 

I would leave the final review, approve and merge for one of the @greenbone/troubadix-maintainers .

Also not sure if it would make sense to split the `test_fail` case into single ones for each case but would also leave this up to the @greenbone/troubadix-maintainers

troubadix/plugins/spaces_before_dots.py
```diff
@@ -0,0 +1,53 @@
...
+    ScriptTag.INSIGHT,
+    ScriptTag.IMPACT,
+    ScriptTag.AFFECTED,
+    ScriptTag.SOLUTION,
+]
+
+# Regex pattern to match:
+# 1. A dot preceded and/or followed by any whitespace character (floating between words)
+# 2. A dot preceded by any whitespace character at the end of the string
+PATTERN = re.compile(r"\s\.\s|\s\.$")
```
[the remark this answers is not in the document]
NiklasHargarter: The '+" can also be left out. 
Changed to the short regex in 1237ff6ae295dbb0c88a388d8d6c288c596b52f7

tests/plugins/test_spaces_before_dots.py
```diff
@@ -0,0 +1,103 @@
...
+        self.assertEqual(len(results), 0)
+
+    def test_fix(self):
+        with self.create_directory() as tempdir:
+            path = tempdir / "file.nasl"
+            content = (
+                """
+                  script_tag(name:"summary", value:"Foo Bar .");
+                  script_tag(name:"vuldetect", value:"Foo Bar .");
+                  script_tag(name:"insight", value:"Foo Bar .");
```
cfi-gb, line 71:
```suggestion
                  script_tag(name:"insight", value:"Foo .NET Bar .");
```

tests/plugins/test_spaces_before_dots.py
```diff
@@ -0,0 +1,103 @@
...
+                  script_tag(name:"impact", value:"Foo . Bar . Foo");
+                """
+                'script_tag(name:"affected", value:"Foo\n.\nBar.");'
+                'script_tag(name:"solution", value:"Foo Bar\n.");'
+            )
+            expected_modified_content = (
+                """
+                  script_tag(name:"summary", value:"Foo Bar.");
+                  script_tag(name:"vuldetect", value:"Foo Bar.");
+                  script_tag(name:"insight", value:"Foo Bar.");
```
cfi-gb, line 81:
```suggestion
                  script_tag(name:"insight", value:"Foo .NET Bar.");
```

For the previous suggestion

tests/plugins/test_spaces_before_dots.py
NiklasHargarter: fixed in f7c8088f9b6cf5074e02fca745abc4eb5dd04f9c

tests/plugins/test_spaces_before_dots.py
NiklasHargarter: fixed in f7c8088f9b6cf5074e02fca745abc4eb5dd04f9c

troubadix/plugins/spaces_before_dots.py
```diff
@@ -0,0 +1,92 @@
...
+            if match:
+                value = match.group("value")
+                value_start = match.start("value")
+
+                for excess_match in PATTERN.finditer(value):
+                    whitespace_pos = excess_match.start() + value_start
+                    self.matches.append((whitespace_pos, excess_match.group()))
+                    fullmatch = match.group()
+                    yield LinterWarning(
+                        f"value of script_tag {match.group('name')} has alteast"
```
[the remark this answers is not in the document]
NiklasHargarter: fixed in f7c8088f9b6cf5074e02fca745abc4eb5dd04f9c

NiklasHargarter:
Wait for confirmation that 22.04 should also be completely excluded.

troubadix/plugins/spaces_before_dots.py
```diff
@@ -0,0 +1,129 @@
...
+                        plugin=self.name,
+                    )
+
+    def fix(self) -> Iterator[LinterResult]:
+
+        if not self.matches:
+            return
+
+        # Sort matches by position, descending order to avoid messing up indices during replacement
+        self.matches.sort(reverse=True, key=itemgetter(0))
```
n-thumann, line 110:
```suggestion
        self.matches.sort(reverse=True)
```
`whitespace_pos` is unique, so that there can only ever be one tuple with the same, right?
In that case, we could drop the `key`, because Python will first compare by the first tuple element and only then consider the second (which won't happen, because the first element is unique anyways).
If there are two elements with the same `whitespace_pos`, itemgetter will (probably) keep the original order, instead of then taking the second tuple element into account.

n-thumann (dismissed):

troubadix/plugins/spaces_before_dots.py
```diff
@@ -0,0 +1,129 @@
...
+]
+
+# Regex pattern to match:
+# 1. A dot preceded and/or followed by any whitespace character (floating between words)
+# 2. A dot preceded by any whitespace character at the end of the string
+PATTERN = re.compile(r"\s+\.(\s|$)")
+IGNORE = [
+    # 21.04 and 22.04 are generated and should not be touched manually
+    "21.04",
+    "22.04",
```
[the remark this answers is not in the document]
NiklasHargarter: done in 0ebb27b978aa83a14db59d3c9552f0834879b433

troubadix/plugins/spaces_before_dots.py
```diff
@@ -0,0 +1,129 @@
...
+        in script_tags that have full sentence values
+        """
+        self.matches = []
+        if nasl_file.suffix == ".inc" or is_ignore_file(nasl_file, IGNORE):
+            return
+        for tag in TAGS:
+            pattern = get_script_tag_pattern(tag)
+            match = pattern.search(file_content)
+            if match:
+                value = match.group("value")
```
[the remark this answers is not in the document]
NiklasHargarter: done in 0ebb27b978aa83a14db59d3c9552f0834879b433

PR #763, opened 2024-10-22
Change: Update validators

n-thumann:
## What
This PR updates the dependency validators.

## Why
We've locked validators to a fixed version in https://github.com/greenbone/troubadix/pull/609, but it's no longer necessary.

## References
Related to https://github.com/greenbone/troubadix/pull/609.

## Checklist

- [ ] Tests

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -308,17 +308,6 @@ files = [
 colorama = {version = "*", markers = "platform_system == \"Windows\""}
 
 [[package]]
-name = "decorator"
-version = "5.1.1"
-description = "Decorators for Humans"
-optional = false
-python-versions = ">=3.5"
-files = [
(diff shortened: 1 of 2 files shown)


cfi-gb:
Note that the following relevant issues / PRs:
- python-validators/validators#296
- python-validators/validators#297

it might be possible that we need to pass `strict_query=True` (defaults to `strict_query=False`) to our code that some of the valid URLs are passing.

n-thumann:
> it might be possible that we need to pass strict_query=True (defaults to strict_query=False) to our code that some of the valid URLs are passing.

Thanks for the hint, I now set `strict_query=False` (I think you mixed up `True` and `False` in your message) and now the unit test with example URL are also fixed.

cfi-gb:
Ah, yes (from https://github.com/python-validators/validators/issues/296#issuecomment-1713231534)

> After applying, you'll have to pass `strict_query = False`, to get those URLs validated.

and from python-validators/validators#297:

> Adds strict_query (default value is True) flag to url validation, functional urls with weird query parts should now pass with strict_query=False

PR #764, opened 2024-10-24
Change: Extend invalid URL plugin and relevant tests.

cfi-gb:
## What
- Adds more cases which should be detected which seems to be not covered by the validators package (yet)
- Excludes a few additional valid / allowed URLs which are now reported as invalid after the version bump of the validators package in #763

## Why
- To catch additional possible malformed URLs introduced by e.g. some automatic extraction or by copy'n'paste like seen / noticed in:
  - greenbone/vulnerability-tests#13985
  - greenbone/vulnerability-tests#13986
  - greenbone/vulnerability-tests#13953
  - greenbone/vulnerability-tests#13959
  - greenbone/vulnerability-tests#14055
- To exclude a few special URLs which seems to be valid but edge cases and reported as invalid by the validators package

## References
None

## Checklist
- [x] Tests

### DIFF
--- a/troubadix/plugins/script_xref_url.py
+++ b/troubadix/plugins/script_xref_url.py
@@ -63,11 +63,67 @@ class CheckScriptXrefUrl(FileContentPlugin):
         )
         for match in matches:
             if match:
-                if match.group("value") not in ALLOWED_URLS and not url(
-                    match.group("value"), strict_query=False
-                ):
+                url_value = match.group("value")
+                if url_value in ALLOWED_URLS:
+                    continue
+
+                full_xref = match.group(0)
+
+                # First a few possible malformed URLs introduced by e.g. some automatic
+                # extraction or by copy'n'paste
+                if url_value.endswith(">") and "<" not in url_value:
                     yield LinterError(
-                        f"{match.group(0)}: Invalid URL value",
+                        f"{full_xref}: Invalid URL value (trailing '>')",
                         file=nasl_file,
                         plugin=self.name,
                     )
+                    continue
+
+                if url_value.endswith(","):
+                    yield LinterError(
+                        f"{full_xref}: Invalid URL value (trailing ',')",
+                        file=nasl_file,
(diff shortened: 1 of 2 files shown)


cfi-gb:
@greenbone/troubadix-maintainers Ready for a first review and (improvement) suggestions always welcome.

troubadix/plugins/script_xref_url.py
```diff
@@ -63,11 +63,65 @@ def check_content(
...
-                        file=nasl_file,
-                        plugin=self.name,
-                    )
+                url_value = match.group("value")
+                full_xref = match.group(0)
+                if url_value not in ALLOWED_URLS:
+
+                    # First a few possible malformed URLs introduced by e.g. some automatic
+                    # extraction or by copy'n'paste
+                    if url_value.endswith(">") and "<" not in url_value:
```
n-thumann, line 72:
```suggestion
                if url_value in ALLOWED_URLS:
                    continue

                # First a few possible malformed URLs introduced by e.g. some automatic
                # extraction or by copy'n'paste
                if url_value.endswith(">") and "<" not in url_value:
```
Now that we have a larger block of code in here, we should return early to keep the indentation low (only included the first few lines in my suggestion to keep the suggestion small)

n-thumann (changes requested):

troubadix/plugins/script_xref_url.py
cfi-gb: Ah, nice catch / hint and indeed makes sense. I have also moved the `full_xref = match.group(0)` after the first continue to only assign this if really required.

n-thumann (approved):

n-thumann:
Forgot to mention this in the review: I ran this plugin against `common/gsf/2024` and there were five errors reported: Three of them newly added by this change and the remaining two caused by a bug in validators (caused by trailing `#` in the URL fragment, fixed by https://github.com/python-validators/validators/pull/405).

PR #776, opened 2024-12-20
Add: update valid_oid plugin for Windows

amy-gb:
## What

Add validation for the generated windows NASL scripts.
Also add missing validation for upcoming  OpenSuse generator.

(The other new generators, HCE and openEuler, do not use NASL so Troubadix is irrelevant)

## Why

This check is currently failing while testing win-vt-generator.

## References

VTA-633

## Checklist

- [x] Tests

### DIFF
--- a/troubadix/plugins/valid_oid.py
+++ b/troubadix/plugins/valid_oid.py
@@ -85,18 +85,17 @@ class CheckValidOID(FileContentPlugin):
         family_pattern = get_special_script_tag_pattern(SpecialScriptTag.FAMILY)
         family_match = family_pattern.search(file_content)
 
+        if family_match is None or family_match.group("value") is None:
+            yield LinterError(
+                "VT is missing a script family!",
+                file=nasl_file,
+                plugin=self.name,
+            )
+            return
+
         # Vendor-specific OIDs
         if "1.3.6.1.4.1.25623.1.1." in oid:
-            if family_match is None or family_match.group("value") is None:
-                yield LinterError(
-                    "VT is missing a script family!",
-                    file=nasl_file,
-                    plugin=self.name,
-                )
-                return
-
             family = family_match.group("value")
-
             vendor_number_match = re.search(
                 r"^1\.3\.6\.1\.4\.1\.25623\.1\.1\.([0-9]+)\.", oid
             )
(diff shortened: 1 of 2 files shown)


troubadix/plugins/valid_oid.py
```diff
@@ -377,6 +405,47 @@ def check_content(
...
 
                 return
+        if "1.3.6.1.4.1.25623.1.3." in oid:
+            family_pattern = get_special_script_tag_pattern(
+                SpecialScriptTag.FAMILY
+            )
+            family_match = family_pattern.search(file_content)
+            if not family_match or not family_match.group("value"):
+                yield LinterError(
+                    "VT is missing a script name!",
```
cfi-gb, line 415: Two notes:
1. This is actually checking for a "script family" and not for a "script name"
2. As the `SpecialScriptTag.FAMILY` is now called /evaluated two times in this plugin would it make sense to move it at the top (e.g. directly after the `if "1.3.6.1.4.1.25623.1." not in oid:` check)?
  - a VT without a `script_family()` tag is invalid anyway
  - Optional: Move the `SpecialScriptTag.NAME` check to the top as well

troubadix/plugins/valid_oid.py
```diff
@@ -81,12 +82,11 @@ def check_content(
...
+        family_pattern = get_special_script_tag_pattern(SpecialScriptTag.FAMILY)
+        family_match = family_pattern.search(file_content)
+
         # Vendor-specific OIDs
         if "1.3.6.1.4.1.25623.1.1." in oid:
-            family_pattern = get_special_script_tag_pattern(
-                SpecialScriptTag.FAMILY
-            )
-            family_match = family_pattern.search(file_content)
             if family_match is None or family_match.group("value") is None:
```
cfi-gb, line 90: Move this outside of the "if" check as well and then drop the duplicated check in the newly added `# Fixed OID-scheme for Windows OIDs` as well?

cfi-gb: So something like this (outside of the `if()`) won't work?

```
        family_pattern = get_special_script_tag_pattern(SpecialScriptTag.FAMILY)
        family_match = family_pattern.search(file_content)
        if family_match is None or family_match.group("value") is None:
            yield LinterError(
                "VT is missing a script family!",
                file=nasl_file,
                plugin=self.name,
            )
            return

        # Vendor-specific OIDs
        if "1.3.6.1.4.1.25623.1.1." in oid:
```

PR #949, opened 2026-04-27
Change: Update the valid OIDs range

iul-gb:
Update the valid OIDs range


troubadix/plugins/valid_oid.py
```diff
@@ -440,7 +440,7 @@ def check_content(
 
         if (
             (10000 <= oid_digit <= 29999)
-            or (50000 <= oid_digit <= 179999)
+            or (50000 <= oid_digit <= 199999)
```
cfi-gb: TBD, as the current / new assigned range is `181NNN`.

mbrinkhoff (approved):

PR #972, opened 2026-07-29
Ruff 16 Migration and dependencies update

amy-gb:
## What

Update the code with new rules
- 90% of the changes are typing related
- notable case of ignoring a new rule: the no_solution plugin has not been using timezones, so i did not add timezones
 

Also update the minimum python version to 3.11.

## Why
passing build

## References

vta-966

## Checklist

- [ ] Tests


troubadix/plugins/valid_oid.py
```diff
@@ -75,7 +75,7 @@ def check_content(
 
         if "1.3.6.1.4.1.25623.1." not in oid:
             yield LinterError(
-                f"script_oid() {invalid_oid} '{str(oid)}'",
+                f"script_oid() {invalid_oid} '{oid!s}'",
```
amy-gb: this string formatting shortcut is new to me 🤔

mbrinkhoff: Same :shrug: Also not sure I like it, it looks a bit too *magic*

troubadix/runner.py
```diff
@@ -135,10 +136,10 @@ def run(self, files: Iterable[Path]) -> bool:
             included=self._included_plugins,
         )
 
-        start = datetime.datetime.now()
+        start = datetime.datetime.now(tz=UTC)
         self._run_pooled(files)
 
-        self._reporter.report_info(f"Time elapsed: {datetime.datetime.now() - start}")
+        self._reporter.report_info(f"Time elapsed: {datetime.datetime.now(tz=UTC) - start}")
```
mbrinkhoff: I am not sure I like this and even would say this is ... more wrong than before ? 

from the [docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.now) about `datetime.now`:

> Return the current local date and time.

As troubadix is mostly a local tool (and whatever TZ is used internally with github runners), I'd say we are more accurate when depending on the simple system provided time over an `UTC` usage because of a linter report.

In this case we dont gain anything (as we take a diff between 2 datetimes from the same machine), generally I would rely on the system time over UTC within troubadix

troubadix/plugins/__init__.py
```diff
@@ -71,24 +72,24 @@ def _is_valid_plugin(cls: Type) -> bool:
 class Plugins:
     def __init__(
         self,
-        file_plugins: Iterable[FilePlugin] = None,
-        files_plugins: Iterable[FilesPlugin] = None,
+        file_plugins: Optional[Iterable[FilePlugin]] = None,
```
mbrinkhoff, line 75: No idea why here its `Optional` but everywhere else it switches to `<Type> | None` over `Optional[<Type>]`

mbrinkhoff (changes requested):

troubadix/runner.py
amy-gb: updated!

troubadix/plugins/valid_oid.py
amy-gb: reverted back to the classic str(), it was only being used in this file

troubadix/plugins/__init__.py
amy-gb: I didn't make one the "rule" so both formats are allowed

mbrinkhoff (approved):

